PR #2, opened 2017-01-24
run tests in travis ci

houqp:
See result from https://travis-ci.org/houqp/iris-api/builds/194836984.

Currently, couple e2e tests are broken due to incomplete dummy data.


jrgp:
Yesss!

PR #4, opened 2017-01-25
add badges in readme

houqp:
also test our first PR travis build


jrgp:
I'm going to merge this. The travis is failing due to the e2etests not quite passing with our dummy data. I'll be working on this.. 

Also, we probably shouldn't have python 2.6 as a test target as we've made incompatible changes for 2.7 (dict/set comprehensions, etc)

PR #7, opened 2017-01-26
Fix oncall api to use events endpoint to get manager roster

jrgp:


src/iris_api/role_lookup/oncall.py
```diff
@@ -42,10 +43,11 @@ def team_members(self, team_name):
         return result
 
     def team_manager(self, team_name):
-        result = self.call_oncall('/teams/%s/rosters/managers' % team_name)
+        now = int(time())
+        result = self.call_oncall('/events?role=manager&team=%s&start__le=%s&end__ge=%s' % (team_name, now, now))
         if not result:
             return None
-        return result['users']
+        return [user['user'] for user in result]
```
dwang159: Suggest renaming the user variable to event, since this endpoint is querying the calendar events rather than users. Also, in case someone has no manager events it may make sense to default to querying for team admins (the previous behavior). Use https://oncall.prod.linkedin.com/api/v0/teams/monitoring-infra/admins to get a list of admin usernames.

PR #12, opened 2017-02-03
Application message quota + rate-limiting support

jrgp:
- Support per-application hard and soft quotas, with message + incident notifications
- New API endpoint for viewing quota settings per app

### DIFF
--- a/db/schema_0.sql
+++ b/db/schema_0.sql
@@ -565,6 +565,31 @@ CREATE TABLE `user_team` (
   CONSTRAINT `user_team_ibfk_1` FOREIGN KEY (`user_id`) REFERENCES `user` (`target_id`),
   CONSTRAINT `user_team_ibfk_2` FOREIGN KEY (`team_id`) REFERENCES `team` (`target_id`)
 ) ENGINE=InnoDB DEFAULT CHARSET=latin1;
+
+
+--
+-- Table structure for table `application_quotas`
+--
+
+DROP TABLE IF EXISTS `application_quotas`;
+CREATE TABLE `application_quotas` (
+  `application_id` int(11) NOT NULL,
+  `hard_quota_threshold` smallint(5) NOT NULL,
+  `soft_quota_threshold` smallint(5) NOT NULL,
+  `hard_quota_duration` smallint(5) NOT NULL,
+  `soft_quota_duration` smallint(5) NOT NULL,
+  `plan_name` varchar(255) NOT NULL,
+  `target_id` bigint(20) NOT NULL,
+  `wait_time` smallint(5) NOT NULL DEFAULT 0,
+  PRIMARY KEY (`application_id`),
+  KEY `application_quotas_plan_name_fk_idx` (`plan_name`),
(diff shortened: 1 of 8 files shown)


db/schema_0.sql
```diff
@@ -565,6 +565,31 @@ CREATE TABLE `user_team` (
   CONSTRAINT `user_team_ibfk_1` FOREIGN KEY (`user_id`) REFERENCES `user` (`target_id`),
   CONSTRAINT `user_team_ibfk_2` FOREIGN KEY (`team_id`) REFERENCES `team` (`target_id`)
 ) ENGINE=InnoDB DEFAULT CHARSET=latin1;
+
+
+--
+-- Table structure for table `application_quotas`
+--
+
+DROP TABLE IF EXISTS `application_quotas`;
```
fellyns: Singular table names are preferred (application_quota vs application_quotas).

src/iris_api/bin/sender.py
```diff
@@ -808,6 +808,14 @@ def distributed_send_message(message):
 
 def fetch_and_send_message():
     message = send_queue.get()
+
+    if not cache.application_quotas(message):
+        logger.warn('Hard message quota exceeded; Dropping this message on floor: %s', message)
+        message_id = message.get('message_id')
+        if message_id:
```
fellyns: We should also log if non persisted messages are being dropped. Also we should add a metric for this.

src/iris_api/sender/cache.py
```diff
@@ -34,20 +38,45 @@
 target_reprioritization = None
 target_names = None
 targets_for_role = None
+application_quotas = None
+
+
+class IrisAuth(requests.auth.AuthBase):
```
fellyns: Since we are the active sender we don't need to call Iris to send a notification or make an incident. Just assume the Iris identity and send the notifications directly. There might be some code dupe/split but it's better than having increased runtime complexity.

src/iris_api/sender/cache.py
```diff
@@ -346,6 +375,166 @@ def __call__(self, message, seen=None):
...
+                                    application, wait_time, last_incident_id)
+                        return
+                except KeyError:
+                    logger.exception('Failed parsing info for incident %s during application %s hard quota breach', last_incident_id, application)
+
+        # Make a new incident
+        try:
+            r = self.iris_client.post('incidents', json={
+                'plan': plan_name,
+                'context': {
```
fellyns: I think the templates would work out better if the metadata were more contained.

example:

{ 'quota-breach': {
    'application': application,
    'limit': limit,
    'duration': duration}
}

That way the data for a given "type" of message is contained within one node and not spread out/sharing multiple nodes.

src/iris_api/bin/sender.py
```diff
@@ -808,6 +808,14 @@ def distributed_send_message(message):
 
 def fetch_and_send_message():
     message = send_queue.get()
+
+    if not cache.application_quotas(message):
```
fellyns: cache seems like a weird namespace for this. It's real SoT data in memory, not cached data.

PR #20, opened 2017-02-14
Add functionality to alter quotas with the concept of ownership

jrgp:
- Add idempotent endpoint for setting quotas for an application

- Add application_owner table which binds people to being owners of an app,
  and these owners are allowed to set quotas. (In future PRs, reset API keys
  and context variables/etc)

### DIFF
--- a/src/iris_api/api.py
+++ b/src/iris_api/api.py
@@ -28,7 +28,8 @@ from . import db
 from . import utils
 from . import cache
 from iris_api.sender import auditlog
-from iris_api.sender.quota import get_application_quotas_query, insert_application_quota_query, required_quota_keys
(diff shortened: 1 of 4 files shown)


src/iris_api/api.py
```diff
@@ -1534,6 +1554,66 @@ def on_get(self, req, resp, app_name):
...
         resp.body = ujson.dumps(quota)
 
+    def on_post(self, req, resp, app_name):
+
+        try:
+            data = ujson.loads(req.context['body'])
+        except ValueError:
+            raise HTTPBadRequest('Invalid json in post body', '')
+
+        required_keys = set(['hard_quota_threshold', 'soft_quota_threshold',
```
houqp: change to a global frozenset?

PR #24, opened 2017-02-18
add /v0/target_roles endpoint

houqp:


src/iris_api/api.py
```diff
@@ -1439,6 +1439,24 @@ def on_post(self, req, resp, username):
...
 
 
+class TargetRoles(object):
+    allow_read_only = False
+
+    def on_get(self, req, resp):
+        session = db.Session()
+        try:
+            sql = '''SELECT `name` FROM `target_role`'''
+            results = session.execute(sql, req.params)
```
jrgp: Do we need req.params here?

houqp: good catch, copy paste error >_<

PR #28, opened 2017-03-02
Add email/sms claim all support

jrgp:
- When 'claim all' submitted through sms or email, claim all incidents currently being sent to that user
- Reply to user with a list of incident IDs that were claimed as a result

### DIFF
--- a/src/iris_api/api.py
+++ b/src/iris_api/api.py
@@ -1855,6 +1855,7 @@ class ResponseMixin(object):
 
         session = db.Session()
         is_batch = False
+        is_claim_all = False
         if isinstance(msg_id, int) or (isinstance(msg_id, basestring) and msg_id.isdigit()):
             # FIXME: return error if message not found for id
             app = get_app_from_msg_id(session, msg_id)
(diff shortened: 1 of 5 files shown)


houqp:
For e2e tests, you can create couple fake incidents before the claim all, then query incidents api to see if they are still active after the claim all command is processed.

src/iris_api/api.py
```diff
@@ -1875,6 +1875,14 @@ def validate_app(app):
             for mid in mid_lst:
                 self.create_response(mid, source, content)
             is_batch = True
+        elif isinstance(msg_id, list) and content == 'claim_all':
+            # Claim all functionality.
+            if not msg_id:
+                raise HTTPBadRequest('Invalid message id', 'No incidents to claim')
+            app = get_app_from_msg_id(session, msg_id[0])
```
fellyns: can/should we assume that all of the claimed incident messages are from one application?

src/iris_api/plugins/core.py
```diff
@@ -79,11 +79,38 @@ def process_iris_batch_claim(self, msg_id, source, mode, cmd, args=None):
...
 
+    def process_claim_all(self, msg_ids, source, mode):
+        owner = utils.lookup_username_from_contact(mode, source)
+        if not owner:
+            return 'Failed to identify owner for these incidents.'
+        if not msg_ids:
+            return 'No messages to claim.'
+        claimed = set()
+        failed_to_claim = set()
+        for msg_id in msg_ids:
```
fellyns: ideally we should do this in bulk in the DB

src/iris_api/utils.py
```diff
@@ -55,6 +55,16 @@ def parse_response(response, mode, source):
                                     LIMIT 1''', {'target_name': target_name}).scalar()
         session.close()
         return msg_id, 'claim'
+    elif re.match('[cC]laim\s+all', response):
```
fellyns: why not make the entire match case insenstive via re.IGNORECASE?

src/iris_api/utils.py
```diff
@@ -55,6 +55,16 @@ def parse_response(response, mode, source):
...
+        session = db.Session()
+        target_name = lookup_username_from_contact(mode, source, session=session)
+        msg_ids = [row[0] for row in session.execute('''SELECT `message`.`id` from `message`
```
fellyns: Do this as part of atomic claiming in the DB instead of fetching a list then trying to claim each one in a CAS fashion to account for race conditions

PR #29, opened 2017-03-03
Base functionality to edit applications

jrgp:
- Support changing variables, sample_context, summary_template, context_template
- Fix dummy data by making one of the dummy users an admin and fixing broken app json


src/iris_api/api.py
```diff
@@ -1552,6 +1552,67 @@ def on_get(self, req, resp, app_name):
...
+            raise HTTPBadRequest('sample_context must be valid json', '')
+
+        if 'context_template' not in data:
+            raise HTTPBadRequest('context_template must be specified', '')
+
+        if 'summary_template' not in data:
+            raise HTTPBadRequest('summary_template must be specified', '')
+
+        new_variables = data.get('variables')
+        if not new_variables or not isinstance(data['variables'], list):
```
houqp: simplify this to `if not isinstance(new_variables, list):`?

src/iris_api/api.py
```diff
@@ -1552,6 +1552,67 @@ def on_get(self, req, resp, app_name):
...
+        session.execute('''UPDATE `application`
+                          SET `context_template` = :context_template, `summary_template` = :summary_template,
+                              `sample_context` = :sample_context
+                          WHERE `id` = :application_id LIMIT 1''', data)
+
+        session.commit()
+        session.close()
+
+        resp.body = '{}'
+        resp.status = HTTP_204
```
houqp: If it's no content, then we should return an empty body right?

src/iris_api/api.py
```diff
@@ -1552,6 +1552,67 @@ def on_get(self, req, resp, app_name):
         resp.status = HTTP_200
         resp.body = ujson.dumps(payload)
 
+    def on_post(self, req, resp, app_name):
```
houqp: I suggest we use PUT for update, POST for creation.

src/iris_api/api.py
jrgp: This breaks the javascript ajax request as jquery expects json. I'll change this to 200 instead.

src/iris_api/api.py
jrgp: Got it

PR #35, opened 2017-03-08
target_role(fix): also include role type in API response

houqp:
required for arbitrary role support from iris frontend

### DIFF
--- a/src/iris_api/api.py
+++ b/src/iris_api/api.py
@@ -1455,11 +1455,40 @@ class TargetRoles(object):
     allow_read_only = False
 
     def on_get(self, req, resp):
+        '''
+        Target role fetch endpoint.
+
+        **Example request**:
+
+        .. sourcecode:: http
+
+           GET /api/v0/target_roles
+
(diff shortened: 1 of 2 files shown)


src/iris_api/api.py
```diff
@@ -1455,11 +1455,40 @@ class TargetRoles(object):
...
+        '''
         session = db.Session()
         try:
-            sql = '''SELECT `name` FROM `target_role`'''
+            sql = '''SELECT `target_role`.`name` AS `name`, `target_type`.`name` AS `type`
+                     FROM `target_role`
+                     JOIN `target_type` on `target_role`.`type_id` = `target_type`.`id`'''
             results = session.execute(sql)
-            payload = ujson.dumps([row for (row,) in results])
+            payload = ujson.dumps([{'name': row[0], 'type': row[1]} for row in results])
```
jrgp: Is there an e2etest for this change?

houqp: test added.

PR #52, opened 2017-03-13
Initial support for twilio delivery status tracking for SMS/CALLS

jrgp:
- Add endpoint for iris-relay to relay delivery updates from twilio

- Create table, with 1:1 mapping to messages, for storing the latest
  delivery update per message

- Add delivery status field to message endpoint. May or may not be null
  depending on whether we've received an update yet

- Pass iris-relay URL to the twilio message create methods

### DIFF
--- a/src/iris_api/api.py
+++ b/src/iris_api/api.py
@@ -115,12 +115,14 @@ single_message_query = '''SELECT `message`.`id` as `id`,
     `mode`.`name` as `mode`,
     `application`.`name` as `application`,
     `priority`.`name` as `priority`,
-    `target`.`name` as `target`
+    `target`.`name` as `target`,
+    `twilio_delivery_status`.`status` as `twilio_delivery_status`
 FROM `message`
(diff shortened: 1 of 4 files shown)


src/iris_api/api.py
```diff
@@ -2437,7 +2465,39 @@ def on_get(self, req, resp, app_name):
                                                                   AND `application_id` = %(application_id)s
                                                                   ORDER BY time_to_claim) as time_to_claim
                                                             WHERE (SELECT @row_id := @row_id + 1)
-                                                            BETWEEN @incident_count/2.0 AND @incident_count/2.0 + 1)'''
+                                                            BETWEEN @incident_count/2.0 AND @incident_count/2.0 + 1)''',
+            'pct_successful_sms_last_month': '''SELECT ROUND((SELECT count(*) FROM `message`
```
houqp: can we rename these stats to associate them with twilio? To make this tracking feature truly optional and make it easy to add other vendor support in the future.

jrgp: Just renamed them to have twilio in the name

PR #56, opened 2017-03-15
Support twilio status updates for calls as well as SMS

jrgp:
The keys are named differently for calls.

Also remove test against 400 as occasionally duplicate sid's are passed which is legal.

### DIFF
--- a/src/iris_api/api.py
+++ b/src/iris_api/api.py
@@ -2236,9 +2236,15 @@ class TwilioDeliveryUpdate(object):
                 'status': post_dict['MessageStatus'],
                 'sid': post_dict['MessageSid']
             }
-        except KeyError as e:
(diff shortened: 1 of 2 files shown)


src/iris_api/api.py
```diff
@@ -2236,9 +2236,15 @@ def on_post(self, req, resp):
                 'status': post_dict['MessageStatus'],
                 'sid': post_dict['MessageSid']
```
houqp: Do you think it's cleaner to avoid the exception handling with dictionary get method?

```python
sid = post_dict.get('MessageSid', post_dict.get('CallSid'))
status = post_dict.get('MessageStatus', post_dict.get('CallStatus'))
if not sid or not status:
    raise
```

jrgp: Yes, actually. Let me fix this.

PR #64, opened 2017-03-17
Ability to send emails from special email addresses to make incidents

jrgp:


src/iris_api/api.py
```diff
@@ -2098,10 +2098,33 @@ def on_post(self, req, resp):
...
+        email_check_result = session.execute('''SELECT `incident_emails`.`application_id`, `plan_active`.`plan_id`
+                                                FROM `incident_emails`
+                                                JOIN `plan_active` ON `plan_active`.`name` = `incident_emails`.`plan_name`
+                                                WHERE `email` = :email''', {'email': source}).fetchone()
+        if email_check_result:
+            incident_info = {
+                'application_id': email_check_result['application_id'],
+                'created': datetime.datetime.utcnow(),
+                'plan_id': email_check_result['plan_id'],
+                'context': ujson.dumps({'body': content})
```
houqp: maybe also add source email address in the context? this can be handy for troubleshooting when multiple emails are using the same plan for the same application.

jrgp: That makes sense. Will implement.

PR #82, opened 2017-03-28
Functionality to view application API keys in the UI

jrgp:
- Add new endpoint that only works for iris admins and iris app owners
- Add logic to display this in the UI


fellyns:
LGTM

PR #98, opened 2017-04-05
Add UI support for deleting and renaming applications

jrgp:
Both actions only available to admins

### DIFF
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -1858,6 +1858,25 @@ class Application(object):
 
         resp.body = '{}'
 
+    def on_delete(self, req, resp, app_name):
+        if not req.context['is_admin']:
+            raise HTTPUnauthorized('Only admins can remove apps', '')
+
+        session = db.Session()
+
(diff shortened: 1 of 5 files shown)


src/iris/api.py
```diff
@@ -1858,6 +1858,26 @@ def on_put(self, req, resp, app_name):
...
+            session.commit()
+        except IntegrityError:
+            raise HTTPBadRequest('Cannot remove app. It has likely already in use.', '')
+        finally:
+            session.close()
+
+        if not affected:
+            raise HTTPBadRequest('No rows changed; app name probably already deleted', '')
+
+        resp.status = HTTP_204
```
houqp: let's keep the status code consistent with the other apis, i.e. return 200 if body is not empty.

jrgp: Fixed this

houqp:
WTF, why are you pushing code at 1:30 am.

jrgp:
Because I can xD

PR #99, opened 2017-04-05
ops: fix docker build & aws provisioning

houqp:
This is ready for merge.


ops/entrypoint.py
```diff
@@ -30,7 +30,12 @@ def load_sqldump(config, sqlfile):
 
 
 def wait_for_mysql(config):
-    db_address = (config['host'], 3306)
+    try:
+        host, port = config['host'].split(':')
+    except:
```
jrgp: hmm can you specify the exceptions that might happen?

PR #103, opened 2017-04-07
Fix path sanitation bug

dwang159:
Avoid unauthorized file reads

### DIFF
--- a/src/iris/ui/__init__.py
+++ b/src/iris/ui/__init__.py
@@ -42,7 +42,7 @@ jinja2_env.assets_environment = assets_env
 jinja2_env.filters['tojson'] = ujson.dumps
 
 mimes = {'.css': 'text/css',
-         '.jpg': 'image/jpg',
+         '.jpg': 'image/jpeg',
          '.js': 'text/javascript',
          '.png': 'image/png',
          '.svg': 'image/svg+xml',
--- a/src/iris/ui/__init__.py
+++ b/src/iris/ui/__init__.py
@@ -111,6 +111,7 @@ class StaticResource(object):
         except IOError:
             raise HTTPNotFound()
 
+
 class Index(object):
     allow_read_only = False
     frontend_route = True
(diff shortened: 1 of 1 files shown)


src/iris/ui/__init__.py
```diff
@@ -69,41 +72,44 @@ def get_local(req, path):
...
-    if req.path.endswith('.woff'):
-        resp.content_type = 'application/font-woff'
-    elif req.path.endswith('.ttf'):
-        resp.content_type = 'application/octet-stream'
-    send_file(req, resp)
+    def __init__(self, path):
+        self.path = path.lstrip('/')
+
+    def on_get(self, req, resp, filename):
+        if req.path.endswith('.js'):
```
jrgp: Instead of the giant if block, can you see if something like this works?

```python
suffix = os.path.splitext(req.path)[1]
mimes = {
	'.js': 'text/javascript',
	# all the rest
}
resp.content_type = mimes.get(suffix)
```

src/iris/ui/__init__.py
```diff
@@ -69,42 +72,44 @@ def get_local(req, path):
...
-    if req.path.endswith('.png'):
-        resp.content_type = 'image/png'
-    elif req.path.endswith('.jpg'):
-        resp.content_type = 'image/jpg'
-    elif req.path.endswith('.svg'):
-        resp.content_type = 'image/svg+xml'
-    send_file(req, resp)
 
+class StaticResource(object):
+    allow_read_only = False
```
jrgp: I think `allow_read_only` should be set to True and `frontend_route` should be set to False, so these static assets are available to the login form.

src/iris/ui/__init__.py
```diff
@@ -38,6 +41,14 @@
 jinja2_env.assets_environment = assets_env
 jinja2_env.filters['tojson'] = ujson.dumps
 
+mimes = {'.css': 'text/css',
+         '.jpg': 'image/jpg',
```
jrgp: Forgot this one. Mimetype for jpg is `image/jpeg`

PR #114, opened 2017-04-11
Add functionality in the UI to edit the default modes per app

jrgp:


src/iris/ui/static/js/iris.js
```diff
@@ -1952,6 +1954,32 @@ iris = {
           self.modelPersist();
           self.render();
       });
+      data.$page.on('submit', data.addDefaultModeForm, function() {
```
shoxv: pass the event in the function here: function(e){

and then e.preventDefault(); at the top of the function

also do this for the other form submits above.

jrgp: Did this

PR #116, opened 2017-04-12
Backend and frontend changes to let app owners change owners

jrgp:
- Previously only iris admins could edit owners

- Prevent an app owner from removing their own access

### DIFF
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -1781,10 +1781,10 @@ class Application(object):
             raise HTTPBadRequest('Application %s not found' % app_name, '')
 
         # Only admins and application owners can change app settings
-        if not req.context['is_admin']:
-            if not session.execute(check_application_ownership_query, {'application_id': app['id'], 'username': req.context['username']}).scalar():
-                session.close()
-                raise HTTPUnauthorized('You don\'t have permissions to update this app\'s quota.', '')
+        is_owner = bool(session.execute(check_application_ownership_query, {'application_id': app['id'], 'username': req.context['username']}).scalar())
+        if not is_owner and not req.context['is_admin']:
+            session.close()
+            raise HTTPUnauthorized('Only admins and app owners can change this app\'s settings.', '')
 
         try:
             ujson.loads(data['sample_context'])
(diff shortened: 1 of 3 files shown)


src/iris/api.py
```diff
@@ -1781,10 +1781,11 @@ def on_put(self, req, resp, app_name):
             raise HTTPBadRequest('Application %s not found' % app_name, '')
 
         # Only admins and application owners can change app settings
-        if not req.context['is_admin']:
-            if not session.execute(check_application_ownership_query, {'application_id': app['id'], 'username': req.context['username']}).scalar():
+        is_owner = bool(session.execute(check_application_ownership_query, {'application_id': app['id'], 'username': req.context['username']}).scalar())
+        if not is_owner:
```
houqp:
```python
if not is_owner and not req.context.get('is_admin'):
```

jrgp: Whoops, good catch. Updated.

src/iris/ui/static/js/iris.js
```diff
@@ -2026,7 +2033,7 @@ iris = {
         app.editable = window.appData.user_admin || owner;
         app.supportViewingKey = window.appData.user_admin || owner;
         app.supportEditingEmailIncidents = window.appData.user_admin || owner;
-        app.showEditOwners = window.appData.user_admin;
+        app.showEditOwners = window.appData.user_admin || owner;
```
shoxv: I'm confused by 'owner' here. Wouldn't `var owner = app.owners.indexOf(window.appData.user) !== -1;` just return true/false? maybe a ternary statement here is what you were going for - var owner = app.owners.indexOf(window.appData.user) !== -1 ? app.owners[window.appData.user] : 'null';

jrgp: Sorry for the variables being named oddly. It would be more appropriate if `owner` was instead named `is_owner` and `user_admin` was instead named `is_admin`.

Its intended for these to be boolean values as they're used in the various if statements in the template to allow certain actions based on whether a user is an owner of this app, or an admin of iris itself.

PR #119, opened 2017-04-13
Tweaks to app stats

jrgp:
- Show fail/success and now other counts for each mode
- Name the metrics in the resulting json dict consistently in a way that
  makes sense and is easier to read
- Also now sort the keys in the resulting dict for clarity reasons
- Use a single query to get all stats for all modes instead of one per each, which
   makes the endpoint take less time

### DIFF
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -3019,70 +3019,6 @@ class ApplicationStats(object):
                                                                   ORDER BY time_to_claim) as time_to_claim
                                                             WHERE (SELECT @row_id := @row_id + 1)
                                                             BETWEEN @incident_count/2.0 AND @incident_count/2.0 + 1)''',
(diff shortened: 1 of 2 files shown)


src/iris/api.py
```diff
@@ -3106,11 +3042,57 @@ def on_get(self, req, resp, app_name):
...
+
+        for mode, status, count in cursor:
+            if isinstance(status, basestring) and status.isdigit():
+                status = int(status)
+            mode_status[mode][status] = count
+
+        for mode in mode_stats_types:
+            status_counts = mode_status[mode]
+            overall = sum(status_counts.values())
+            fail_count = float(sum(status_counts.pop(key, 0) for key in mode_stats_types[mode]['fail']))
```
houqp: Looks like you can skip all the remaining calculation if over is None.

jrgp: Fixed

PR #121, opened 2017-04-14
doc: auto generate rest api spec based on registered routes

houqp:


src/iris/api.py
```diff
@@ -3129,78 +3195,89 @@ def json_error_serializer(req, resp, exception):
...
+    if config['server'].get('disable_auth'):
+        debug = True
+    master_sender = config['sender'].get('master_sender', config['sender'])
+    master_sender_addr = (master_sender['host'], master_sender['port'])
+    # all notifications go through master sender for now
+    app = construct_falcon_api(
+        debug, healthcheck_path, iris_sender_app, master_sender_addr)
 
     # Need to call this after all routes have been created
     app = ui.init(config, app)
```
jrgp: Shouldn't the `ui.init()` be at the end of `construct_falcon_api()`?

Or are is it not there to avoid having the UI routes added to docs?

houqp: To avoid UI routes. That's why I used different names to distinguish them: `app` = `api` + `ui`.

PR #125, opened 2017-04-15
ci: fail early on error in install.sh

houqp:
also get rid of the unnecessary virtualenv setup since travis build already runs in a virtualenv.


jrgp:
Nice

PR #126, opened 2017-04-18
Add CORS

dwang159:
Needed for integration with Oncall (getting plan names)


src/iris/api.py
```diff
@@ -3257,7 +3261,7 @@ def get_api(config):
     master_sender_addr = (master_sender['host'], master_sender['port'])
     # all notifications go through master sender for now
     app = construct_falcon_api(
```
houqp: https://github.com/linkedin/iris/blob/master/src/iris/doc_helper.py needs to be updated too otherwise the document build will fail.

dwang159: Added [ ] as allowed origins list in doc_helper.py

PR #127, opened 2017-04-18
Fix messages never being marked as sent, and clarify logging

jrgp:
- Update the message subject/body separately from the other message metadata,
  to avoid the sent/active status not being marked when we fail to update the
  row due to a giant subject or body

- This will fix 1) messages not being marked as sent which breaks aggregation and
   sender queueing and 2) the message endpoint not
   working due to the mode join not being satisfied.

- When either fail to update, log the application name as well as the message ID

- Now truncate the message body from 65k to 10k for good measure

- Also additional logging for when we drop OOB messages

### DIFF
--- a/src/iris/bin/sender.py
+++ b/src/iris/bin/sender.py
@@ -168,6 +168,9 @@ UPDATE_MESSAGE_BODY_SQL = '''UPDATE `message`
 
 PRUNE_OLD_AUDIT_LOGS_SQL = '''DELETE FROM `message_changelog` WHERE `date` < DATE_SUB(CURDATE(), INTERVAL 3 MONTH)'''
 
+# When a rendered message body is longer than this number of characters, drop it.
+MAX_MESSAGE_BODY_LENGTH = 10000
+
 # logging
 logger = logging.getLogger()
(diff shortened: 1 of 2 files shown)


src/iris/bin/sender.py
```diff
@@ -755,37 +756,53 @@ def render(message):
...
+        connection.commit()
+    except DataError:
+        logger.exception('Failed updating message metadata status (message ID %s) (application %s)', message.get('message_id', '?'), message.get('application', '?'))
+
+    # Update subject and body separately, as they may fail and we don't necessarily care if they do
     if len(message['subject']) > 255:
         message['subject'] = message['subject'][:255]
-    if len(message['body']) > 65000:
+
+    if len(message['body']) > 10000:
```
houqp: Would it be better to do this check right after https://github.com/linkedin/iris/blob/master/src/iris/bin/sender.py#L903? Then we can skip the whole `distributed_send_message` call.

Problem with doing it after `distributed_send_message` is the user might receive the message delivered but it would be missing from the DB.

jrgp: I added this check after the render() call so it gets run there as well and prevents the message from being sent. Still need to run it here as this function is called in a few other places.

PR #136, opened 2017-04-21
doc: add security section

houqp:


docs/source/iris.rst
```diff
@@ -28,3 +28,14 @@ As can be seen, Iris is the crux of the entire framework:
...
+
+
+Security
+--------
+
+Since Iris will be integrated with internal infrastructure, please do not expose
+Iris web and API to the public internet.
+
+Please also make sure **only trusted** users are given access to create and
+update templates since those templates will be renedered using Jinja, whose
```
jrgp: rendered seems to be mispelled. Maybe say something like the jinja's sandbox isn't good enough to completely prevent running bad python code?

PR #143, opened 2017-04-25
Support overriding config from env vars.

houqp:


ops/packer/iris.yaml
```diff
@@ -97,7 +97,7 @@ provisioners:
       - sudo -Hu iris ln -s /home/iris/source/db /home/iris/db
       - sudo -Hu iris mkdir /home/iris/config
       - sudo -Hu iris virtualenv /home/iris/env
-      - sudo -Hu iris /bin/bash -c 'source /home/iris/env/bin/activate && cd /home/iris/source && python setup.py install'
+      - sudo -Hu iris /bin/bash -c 'source /home/iris/env/bin/activate && cd /home/iris/source && pip install ".[ldap,prometheus]"'
```
jrgp: Will changing away from `python setup.py install` cause setup.py's `console_script` `entry_points` not be created anymore?

Or are they not needed here? Eg this image isn't used for running sender.

houqp: `pip install .` also installs those entry points, but i just noticed those extra install flags are not working without `-e`. Looking into it now.

houqp: nvm, pip install is working as expected

PR #144, opened 2017-04-25
Add application plans endpoint

dwang159:
Add resource allowing GET that returns active plans that support a given application.


src/iris/api.py
```diff
@@ -2299,6 +2299,36 @@ def on_put(self, req, resp, app_name):
...
+                   JOIN `template_content` ON `template`.`id` = `template_content`.`template_id`
+                   JOIN `application` ON `template_content`.`application_id` = `application`.`id`
+                   JOIN `target` ON `target`.`id` = `plan`.`user_id`
+                   WHERE `application`.`name` = %%s
+                   GROUP BY `plan`.`id`''' % ','.join(plan_columns[f] for f in fields)
+
+        connection = db.engine.raw_connection()
+        cursor = connection.cursor(db.dict_cursor)
+        cursor.execute(query, app_name)
+        print query % app_name
```
jrgp: Hmmm print statement left over from debugging?

src/iris/api.py
```diff
@@ -2299,6 +2299,36 @@ def on_put(self, req, resp, app_name):
...
+        payload = cursor.fetchall()
+        cursor.close()
+        connection.close()
+        resp.body = ujson.dumps(payload)
```
jrgp: Hmm `resp.body = ujson.dumps(cursor)` might also work and be faster

PR #147, opened 2017-04-25
Get rid of some 5xxs, update ApplicationPlans

dwang159:
Fix 5xxs on invalid query params and invalid fields. Add name filtering and docs to ApplicationPlans

### DIFF
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -878,8 +878,9 @@ class Plans(object):
         query_limit = req.get_param_as_int('limit')
         req.params.pop('limit', None)
         fields = req.get_param_as_list('fields')
+        fields = [f for f in fields if f in plan_columns]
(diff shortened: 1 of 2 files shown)


src/iris/api.py
```diff
@@ -1319,9 +1325,10 @@ class Messages(object):
 
     def on_get(self, req, resp):
         fields = req.get_param_as_list('fields')
-        if fields is None:
-            fields = message_columns
         req.params.pop('fields', None)
+        fields = [f for f in fields if f in message_columns]
```
jrgp: Why not do something similar to this instead?

`fields = set(fields) & set(message_columns)`

That will get only the columns present in both and is likely more efficient than the list comprehension that iterates through fields while checking for presence in message_columns.

PR #153, opened 2017-04-28
support message body preview for slack

houqp:
Other minor changes include:

* make sender wait for API to go online on init
* don't crash sender on role lookup exception
* deprecate `role_lookup` config
* add Procfile for goreman

### DIFF
--- a/src/iris/role_lookup/__init__.py
+++ b/src/iris/role_lookup/__init__.py
@@ -2,15 +2,24 @@
 # See LICENSE in the project root for license information.
 
 from iris.custom_import import import_custom_module
+import logging
+logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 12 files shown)


src/iris/role_lookup/__init__.py
```diff
@@ -2,15 +2,24 @@
...
+        modules = ['user', 'mailing_list']
+
+    imported_modules = []
+    for m in modules:
+        try:
+            imported_modules.append(
+                import_custom_module('iris.role_lookup', m)(config))
+            logger.info('Loaded lookup modules: %s', m)
+        except:
+            logger.error('Failed to load role lookup module: %s', m)
```
jrgp: Maybe do `logger.exception` instead, so the user can see why the load failed?

PR #173, opened 2017-05-03
Add details of the hmac auth header to docs

jrgp:


kripplek:
That explains the error in my Lib. ty

PR #182, opened 2017-05-05
use gunicorn for run_server

houqp:


src/iris/bin/run_server.py
```diff
@@ -3,83 +3,42 @@
...
-            time.sleep(3)
-            print 'Restart.'
-            os.execvp(sys.argv[0], sys.argv)
+    def __init__(self, app, options=None):
+        self.options = options or {}
+        self.application = app
+        super(StandaloneApplication, self).__init__()
 
+    def load_config(self):
+        config = dict([(key, value) for key, value in iteritems(self.options)
```
jrgp: why not use a dict comprehension instead?

PR #183, opened 2017-05-06
return 400 on invalid type for query values

houqp:


src/iris/api.py
```diff
@@ -583,14 +583,18 @@ def gen_where_filter_clause(connection, filters, filter_types, kwargs):
...
+                        op = 'eq'
+                        val = col_type(values[0])
+                    else:
+                        val = tuple([col_type(v) for v in values])
                 else:
-                    val = tuple([col_type(v) for v in values])
-            else:
-                val = col_type(val)
+                    val = col_type(val)
+            except Exception:
```
jrgp: Why not specifically catch ValueError/TypeError?

PR #188, opened 2017-05-15
Add streamlabs "trusted by" logo

jrgp:


index.html
```diff
@@ -196,6 +196,9 @@ <h3 class="left">
                                 </svg>
                             </a>
                         </li>
+                        <li>
+                          <a href="https://streamlabs.com"><img src="img/streamlabs.png" width="150" height="26"></a>
```
shoxv: can you add an alt-tag to the image. also we shouldnt assign a height because it will distort the image. just the width will be good. probably bump it to 200px to make it a similar sizing to the LI one.

jrgp: Updated

PR #205, opened 2017-05-31
chore: reduce sender log volume

houqp:


src/iris/sender/rpc.py
```diff
@@ -142,9 +142,9 @@ def handle_api_notification_request(socket, address, req):
                     notification['application'])
         return
 
-    logger.info('-> %s OK, to %s:%s (%s:%s)',
-                address, role, target, notification['application'],
-                notification.get('priority', notification.get('mode', '?')))
+    logger.debug('-> %s OK, to %s:%s (%s:%s)',
```
jrgp: While this makes sense, won't it make it a bit harder to determine which applications are sending the most to which users? In the past weeks the `'-> %s OK, to %s:%s (%s:%s)'` messages has made diagnosing that easy.

Maybe this should be left as is with the other info's converted to debug's?

PR #210, opened 2017-05-31
Fix buggy behavior on no contacts

dwang159:
Before, sender would silently mark messages as changed from the
fallback mode to invalid if a mode with no contact info was
specified.
Now, sender attempts to change mode to fallback mode, then
increments task failure metric if that is also unsuccessful.

### DIFF
--- a/src/iris/bin/sender.py
+++ b/src/iris/bin/sender.py
@@ -593,9 +593,14 @@ def set_target_fallback_mode(message):
         cursor.close()
         connection.close()
 
+        old_mode = message['mode']
         message['destination'] = destination
         message['mode'] = mode
         message['mode_id'] = mode_id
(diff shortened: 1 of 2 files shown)


src/iris/bin/sender.py
```diff
@@ -696,10 +702,13 @@ def set_target_contact(message):
...
         if result:
             cache.target_reprioritization(message)
+        else:
+            logger.warn('target does not have mode %r', message)
+            result = set_target_fallback_mode(message)
         return result
     except (ValueError, TypeError):
         logger.exception('target does not have mode %r', message)
-        return set_target_fallback_mode(message)
+        result = set_target_fallback_mode(message)
```
houqp: what does it get us by setting the result here?

PR #226, opened 2017-06-14
Automatic sender leader election and failover.

jrgp:
- Move slave logic from rpc to new `Coordinator` class
- That `Coordinator` class periodically keeps track of the current master as well as slaves
- Sender now refers to that new class to determine, in realtime, whether or not it's the master and should be doing masterly things.
- Add two new tables:

1) `sender_master_election`: Single row used to keep track of master sender
2) `sender_instances`: One row per sender that exists, so we can determine slaves

### DIFF
--- a/src/iris/bin/sender.py
+++ b/src/iris/bin/sender.py
@@ -1118,6 +1118,8 @@ def main():
     global config
     config = load_config()
 
+    start_time = time.time()
+
     logger.info('[-] bootstraping sender...')
     init_sender(config)
     init_plugins(config.get('plugins', {}))
@@ -1194,9 +1196,11 @@ def main():
         for i in bad_workers:
             worker_tasks[i] = spawn(worker)
 
+        now = time.time()
+        metrics.set('sender_uptime', int(now - start_time))
+
         spawn(metrics.emit)
 
-        now = time.time()
         elapsed_time = now - runtime
         nap_time = max(0, interval - elapsed_time)
         logger.info('--> sender loop finished in %s seconds - sleeping %s seconds',
--- a/src/iris/bin/sender.py
+++ b/src/iris/bin/sender.py
@@ -8,6 +8,7 @@ import logging
 import time
 import ujson
 import os
+import socket
 
 from collections import defaultdict
 from iris.plugins import init_plugins
@@ -23,6 +24,7 @@ from iris.sender.message import update_message_mode
 from iris.sender.oneclick import oneclick_email_markup, generate_oneclick_url
 from iris import cache as api_cache
 from iris.sender.quota import ApplicationQuota
+from iris.sender.coordinator import Coordinator
 from pymysql import DataError, IntegrityError
 # queue for sending messages
 from iris.sender.shared import send_queue, add_mode_stat
@@ -210,6 +212,10 @@ message_queue = queue.Queue()
 
 # Quota object used for rate limiting
 quota = None
+
+# Coordinator object for sender master election
+coordinator = None
+
 
 default_sender_metrics = {
     'email_cnt': 0, 'email_total': 0, 'email_fail': 0, 'email_sent': 0, 'email_max': 0,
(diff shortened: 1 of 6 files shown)


src/iris/sender/coordinator.py
```diff
@@ -0,0 +1,106 @@
...
+
+        # Also keep track of us in the list of online senders
+        session.execute(UPDATE_INSTANCES_QUERY, {'me': self.me})
+        session.commit()
+
+        # Keep track of the actual master locally
+        self.master = session.execute(GET_MASTER_QUERY).scalar()
+
+        # Record locally whether we're the master or not
+        self.is_master = self.master == self.me
```
houqp: skip the rest if it's not a master?

src/iris/sender/coordinator.py
```diff
@@ -0,0 +1,106 @@
...
+        self.me = '%s:%s' % (hostname, port)
+        self.is_master = None
+        self.master = None
+
+    def am_i_master(self):
+        return self.is_master
+
+    def get_current_master(self):
+        session = db.Session()
+        master = session.execute(GET_MASTER_QUERY).scalar()
```
houqp: if we want to keep `self.master` always up to date, then we should update it here too. If not, we can just remove `self.master` altogether.

houqp: If we want to keep `self.master`, then I suggest we remove `self.is_master`, so it's one less variable that we need to keep it up to date. We can replace all `self.is_master` with `self.me == self.master`.

jrgp: i thought about this, but isn't the string comparison `self.me == self.master` more expensive than `self.is_master == True ` ?

src/iris/sender/coordinator.py
jrgp: good idea

src/iris/sender/coordinator.py
jrgp: Removed `self.master` in favor of `self.is_master` which is calculated at update time.

src/iris/bin/sender.py
```diff
@@ -1120,7 +1145,18 @@ def main():
         cache.refresh()
         cache.purge()
 
-        if is_master:
+        # If we're currently a master, ensure our master-greenlets are running
+        # and we're doing the master duties
+        if coordinator.am_i_master():
+            if not bool(gwatch_renewer_task):
+                if should_mock_gwatch_renewer:
```
fellyns: instead of checking on if we should mock or not every loop have the main code check and swap out the function  with the mocked function.

src/iris/sender/coordinator.py
```diff
@@ -0,0 +1,158 @@
...
+REMOVE_OLD_INSTANCES_FREQUENCY = 3600
+REMOVE_OLD_INSTANCES_TIMEOUT = 60
+
+GET_MASTER_QUERY = '''SELECT `sender_address` FROM `sender_master_election` WHERE `anchor` = 1'''
+
+GET_SLAVES_QUERY = '''
+  SELECT `sender_address`
+  FROM `sender_instances`
+  WHERE `last_seen` > NOW() - INTERVAL CAST(:timeout AS UNSIGNED) SECOND
+  AND `sender_address` NOT IN (%s)''' % GET_MASTER_QUERY
```
fellyns: use a join here instead of dynamically building the sql up with a subselect

src/iris/sender/coordinator.py
```diff
@@ -0,0 +1,158 @@
...
+
+UPDATE_MASTER_QUERY = '''
```
fellyns: I am interested in how well this tolerates the Galera WAN cluster - it looks potentially deadlocky especially in the case of clock drift on the DB nodes.

src/iris/sender/coordinator.py
```diff
@@ -0,0 +1,158 @@
...
+        session = db.Session()
+        slaves = []
+        for row in session.execute(GET_SLAVES_QUERY, {'timeout': SENDER_ALIVE_TIMEOUT}):
+            address = self.address_to_tuple(row[0])
+            if address:
+                slaves.append(address)
+        session.close()
+        return slaves
+
+    def address_to_tuple(self, address):
```
fellyns: this function can throw an exception but no where is that exception caught - is it acceptable to blow up here?

src/iris/sender/coordinator.py
```diff
@@ -0,0 +1,158 @@
...
+        try:
+            host, port = address.split(':', 1)
```
fellyns: Do we expect more than one colon? - if not why limit the split instead of catching the ValueError from too many values to unpack?

jrgp: Fixed this.

src/iris/sender/coordinator.py
jrgp: Fixed this

src/iris/sender/coordinator.py
jrgp: Fixed to use join

PR #228, opened 2017-06-16
Fix reported unicode exceptions with OOB messages

jrgp:


src/iris/bin/sender.py
```diff
@@ -746,11 +747,17 @@ def render(message):
...
+                        except UnicodeDecodeError:
+                            message['subject'] = mode_template['subject'].render(**sanitize_unicode_dict(message['context']))
                     except Exception as e:
                         error = 'template %(template)s - %(application)s - %(mode)s - subject failed to render: ' + str(e)
                     try:
-                        message['body'] += mode_template['body'].render(**message['context'])
+                        try:
+                            message['body'] += mode_template['body'].render(**message['context'])
+                        except UnicodeDecodeError:
+                            message['subject'] = mode_template['body'].render(**sanitize_unicode_dict(message['context']))
```
houqp: `message['subject'] = mode_template['body']`, expected?

jrgp: Nope. WIll fix that.

PR #230, opened 2017-06-22
(Optionally) Use ZooKeeper instead of MySQL for sender election

jrgp:
- Use ZK with Kazoo instead of MySQL as we had problems with our Galera setup

- Catch kill signals to immediately remove from cluster

- Bring back support for hard coding in master and slave configuration in
  config files. This new ZK support is 100% optional and does not give us a hard
  dependency on ZK.

- Add new NonClusterCoordinator class to provide for hard coded cluster configs
  easily

### DIFF
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -1477,9 +1477,13 @@ class Notifications(object):
     allow_read_no_auth = False
     required_attrs = frozenset(['target', 'role', 'subject'])
 
-    def __init__(self, default_sender_addr):
+    def __init__(self, zk_hosts, default_sender_addr):
         self.default_sender_addr = default_sender_addr
-        self.coordination = Coordinator(None, None)
+        if zk_hosts:
+            self.coordinator = Coordinator(zk_hosts=zk_hosts, hostname=None, port=None, join_cluster=False)
+        else:
+            logger.info('Not using ZK to get senders. Using host %s for master instead.', default_sender_addr)
+            self.coordinator = None
 
     def on_post(self, req, resp):
         '''
@@ -1586,12 +1590,16 @@ class Notifications(object):
 
         message['application'] = req.context['app']['name']
 
-        sender_addr = self.coordination.get_current_master()
-        if sender_addr:
-            logger.info('Relaying message to current master sender: %s', sender_addr)
+        # If we're using ZK, try that to get master
+        if self.coordinator:
+            sender_addr = self.coordinator.get_current_master()
+            if sender_addr:
+                logger.info('Relaying message to current master sender: %s', sender_addr)
+            else:
+                sender_addr = self.default_sender_addr
+                logger.error('Failed getting current sender master. Falling back to %s', sender_addr)
         else:
             sender_addr = self.default_sender_addr
-            logger.error('Failed getting current sender master. Falling back to %s', sender_addr)
 
         s = None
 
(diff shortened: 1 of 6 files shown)


houqp:
so production percona test didn't go well?

jrgp:
@houqp it didn't work out in prod. Master status kept being switched through nodes even though none were down.

src/iris/api.py
```diff
@@ -3684,11 +3693,14 @@ def get_api(config):
     debug = False
     if config['server'].get('disable_auth'):
         debug = True
-    master_sender = config['sender'].get('master_sender', config['sender'])
-    master_sender_addr = (master_sender['host'], master_sender['port'])
+
+    default_master_sender = config['sender'].get('master_sender', config['sender'])
```
fellyns: Can we avoid .get() tests on the config by ensuring that it is always there when generating the config dictionary?

src/iris/bin/sender.py
```diff
@@ -1084,7 +1086,22 @@ def mock_gwatch_renewer():
...
 
+def sender_shutdown():
+    logger.info('Shutting server..')
+
+    # Immediately release all locks and give up any master status and slave presence
+    if coordinator:
+        coordinator.leave_cluster()
+
+    # Force quit. Avoid sender process existing longer than it needs to
+    os._exit(0)
```
fellyns: Will this allow us to finish sending anything we are in the process of sending?

src/iris/sender/coordinator.py
```diff
@@ -146,40 +103,41 @@ def update_forever(self):
...
+
+class NonClusterCoordinator():
+    def __init__(self, is_master, slaves=[]):
+        self.is_master = is_master
+        if is_master:
+            logger.info('I am the master sender')
+        else:
+            logger.info('I am a slave sender')
+
+        self.slaves = cycle([(slave['host'], slave['port']) for slave in slaves])
```
fellyns: you don't need the list comp here, a generator expression will do

self.slaves = cycle((slave['host'], slave['port']) for slave in slaves)

jrgp: Fixed this

src/iris/bin/sender.py
jrgp: Currently, this will not save any existing OOB messages in the queue. It shouldn't affect messages coming from incidents though as another sender will take on continuing incident steps.

src/iris/api.py
```diff
@@ -3607,7 +3616,7 @@ def json_error_serializer(req, resp, exception):
     resp.content_type = 'application/json'
 
 
-def construct_falcon_api(debug, healthcheck_path, allowed_origins, iris_sender_app, sender_addr):
+def construct_falcon_api(debug, healthcheck_path, allowed_origins, iris_sender_app, zk_hosts, default_sender_addr):
```
jrgp: That file and those functions could use a refactor, and that will be addressed in a later PR.

src/iris/api.py
jrgp: In this case, we're trying to get a master_sender subkey, but that might not exist, so we're defaulting to the parent sender key which should always exist. This can be made better in the future though.

jrgp:
Github wouldnt let me update this PR so I made this new PR instead. https://github.com/linkedin/iris/pull/231

PR #233, opened 2017-06-27
Guarded session refactor

dwang159:
move session to db, change utils to use guarded session and/or clean up
sessions properly.


jrgp:
Looks like tests fail because flake8:

    src/iris/db.py:47:14: W292 no newline at end of file

PR #241, opened 2017-07-09
fix uwsgi config for docker image

houqp:


jrgp:
I don't think kazoo is an optional dependency because when `from iris.sender.coordinator import NonClusterCoordinator` runs, coordinator will still try to import the kazoo libs.

src/iris/api.py
```diff
@@ -3,7 +3,10 @@
 
 from __future__ import absolute_import
 
-from gevent import socket
+from gevent import monkey
+monkey.patch_all()  # NOQA
```
jrgp: Is this needed? I thought when you use uwsgi or gunicorn with gevent it automatically monkey patches everything.

houqp:
good catch :+1:  updated the PR.

PR #248, opened 2017-08-04
Add support for dynamic plans

dwang159:
Let users create plans with target numbers rather than specific
targets, then define mappings for those number on incident creation

### DIFF
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -276,10 +276,11 @@ single_plan_query_steps = '''SELECT `plan_notification`.`id` as `id`,
     `target_role`.`name` as `role`,
     `target`.`name` as `target`,
     `plan_notification`.`template` as `template`,
-    `priority`.`name` as `priority`
+    `priority`.`name` as `priority`,
+    `plan_notification`.`dynamic_index` AS `dynamic_index`
 FROM `plan_notification`
-JOIN `target` ON `plan_notification`.`target_id` = `target`.`id`
-JOIN `target_role` ON `plan_notification`.`role_id` = `target_role`.`id`
+LEFT OUTER JOIN `target` ON `plan_notification`.`target_id` = `target`.`id`
+LEFT OUTER JOIN `target_role` ON `plan_notification`.`role_id` = `target_role`.`id`
 JOIN `priority` ON `plan_notification`.`priority_id` = `priority`.`id`
 WHERE `plan_notification`.`plan_id` = %s
 ORDER BY `plan_notification`.`step`'''
@@ -365,6 +366,18 @@ insert_plan_step_query = '''INSERT INTO `plan_notification` (
     :role_id,
     :repeat,
     :wait
+)'''
+
+insert_dynamic_step_query = '''INSERT INTO `plan_notification` (
+    `plan_id`, `step`, `priority_id`, `template`, `repeat`, `wait`, `dynamic_index`
+) VALUES (
+    :plan_id,
+    :step,
+    :priority_id,
+    :template,
+    :repeat,
+    :wait,
+    :dynamic_index
 )'''
 
 reprioritization_setting_query = '''SELECT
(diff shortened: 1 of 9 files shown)


src/iris/api.py
```diff
@@ -1204,6 +1233,24 @@ def on_post(self, req, resp):
...
                     raise HTTPBadRequest('Invalid application')
+            if 'dynamic_targets' in incident_params:
+                num_targets = session.execute('SELECT COUNT(DISTINCT `dynamic_index`) FROM `plan_notification` '
+                                              'WHERE `plan_id` = :plan_id',
+                                              {'plan_id': plan_id}).scalar()
+                if num_targets != len(incident_params['dynamic_targets']):
+                    raise HTTPBadRequest('Invalid number of dynamic targets')
+
+                for dynamic_target in incident_params.get('dynamic_targets', []):
+                    target = session.execute('''SELECT `target_role`.`id` AS `role_id`, `target`.`id` AS `target_id`
```
jrgp: i think this could be shortened to `session.execute().scalar()`, which would be None if there was no rows returned, then you can check that instead of rowcount

src/iris/ui/static/js/iris.js
```diff
@@ -129,6 +129,7 @@ iris = {
       notificationTemplateSource: $('#plan-notification-template').html(),
       stepTemplateSource: $('#plan-step-template').html(),
       trackingTemplateSource: $('#plan-tracking-notification-template').html(),
+      testPlanSource: $('#test-plan-template').html(),
```
jrgp: maybe add `Template` to this variable name to keep it consistent with the others

src/iris/ui/static/js/iris.js
```diff
@@ -266,8 +274,16 @@ iris = {
             for (i = 0; i < response.steps.length; i++) {
               for (j = 0; j < response.steps[i].length; j++) {
                 response.steps[i][j].count = response.steps[i][j].repeat + 1;
+                if (response.steps[i][j].dynamic_index !== null) {
+                  max_idx = Math.max(max_idx, response.steps[i][j].dynamic_index);
+                }
+                response.steps[i][j].dynamic = response.steps[i][j].dynamic_index !== null
```
jrgp: semicolon at end of line

src/iris/ui/static/js/iris.js
```diff
@@ -408,9 +456,23 @@ iris = {
...
     addNotification: function(event){
       var $step = $(event.target).parents('.plan-step'),
+          notifications = $('.plan-notification');
           template = Handlebars.compile(this.data.notificationTemplateSource);
-
-      $step.find('.plan-notification-add').before(template(this.data.blankModel));
+          max_id = 0;
+
+      for (i = 0; i < notifications.length; i++){
+        id = notifications[i].id;
```
jrgp: might want to add `var` here to keep id's scope to the for loop's body

src/iris/ui/static/js/iris.js
```diff
@@ -408,9 +456,23 @@ iris = {
     },
     addNotification: function(event){
       var $step = $(event.target).parents('.plan-step'),
+          notifications = $('.plan-notification');
```
shoxv: the end of this line should be a comma not semicolon. this is causing 'template' to become a global variable.

src/iris/api.py
dwang159: Changed to `fetchone()` (since I'm grabbing 2 values)

src/iris/ui/static/js/iris.js
```diff
@@ -408,9 +456,23 @@ iris = {
...
       var $step = $(event.target).parents('.plan-step'),
-          template = Handlebars.compile(this.data.notificationTemplateSource);
-
-      $step.find('.plan-notification-add').before(template(this.data.blankModel));
+          notifications = $('.plan-notification'),
+          template = Handlebars.compile(this.data.notificationTemplateSource),
+          max_id = 0;
+
+      for (var i = 0; i < notifications.length; i++){
+        id = notifications[i].id;
```
jrgp: I think `id` is still scoped globally. (You fixed `i`)

dwang159: Ran some JS code inspection. Should be good now.

PR #257, opened 2017-08-17
Split inactive incident sql to avoid deadlock

dwang159:


src/iris/bin/sender.py
```diff
@@ -351,9 +352,12 @@ def deactivate():
     # this deadlocks sometimes. try until it doesn't.
     for i in xrange(max_retries):
         try:
-            cursor.execute(INACTIVE_SQL)
-            connection.commit()
-            break
+            cursor.execute(GET_INACTIVE_IDS_SQL)
+            ids = tuple(r[0] for r in cursor)
+            if len(ids) > 0:
```
jrgp: I think you could do `if ids` instead of `if len(ids) > 0` to check for the empty tuple

PR #258, opened 2017-08-18
(WIP) Utilize a separate history table for (un)claiming incidents

jrgp:
**WIP because performance is very bad when tested against prod dataset.**

- Avoid `UPDATE`'ing the `active` and `owner_id` columns of the `incident`
  table by the API

- Instead, insert either "claim" or "unclaim" actions into a history
  table

- Show the claim/unclaim actions for a given incident on the incident
  view page


jrgp:
Closing until I can fix the perf here.

PR #259, opened 2017-08-18
Add retry to failed Twilio messages

dwang159:


test/e2etest.py
```diff
@@ -2636,6 +2637,52 @@ def test_twilio_delivery_update(fake_message_id):
...
+    if not fake_message_id:
+        pytest.skip('We do not have enough data in DB to do this test')
+
+    message_sid = uuid.uuid4().hex
+
+    with iris_ctl.db_from_config(sample_db_config) as (conn, cursor):
+        cursor.execute('''DELETE FROM `twilio_delivery_status` WHERE `message_id` = %s''', fake_message_id)
+        cursor.execute('''INSERT INTO `twilio_delivery_status` (`twilio_sid`, `message_id`)
+                          VALUES (%s, %s)''', (message_sid, fake_message_id))
+        cursor.execute('''DELETE FROM `twilio_retry` WHERE `retry_id` = %s OR `message_id` = %s''',
```
jrgp: you could do `%(id)s` and `{'id': fake_message_id}` to avoid specifying it twice.

db/schema_0.sql
```diff
@@ -657,6 +657,17 @@ CREATE TABLE `twilio_delivery_status` (
   CONSTRAINT `twilio_delivery_status_message_id_ibfk` FOREIGN KEY (`message_id`) REFERENCES `message` (`id`) ON DELETE CASCADE ON UPDATE CASCADE
 ) ENGINE=InnoDB DEFAULT CHARSET=latin1;
 
+
+DROP TABLE IF EXISTS `twilio_retry`;
+CREATE TABLE `twilio_retry` (
+  `message_id` bigint(20) NOT NULL,
+  `retry_id` bigint(20) NOT NULL,
+  PRIMARY KEY (`message_id`),
```
jrgp: I'm wondering if you could make this use a composite primary key of the two columns, instead of PK'ing message_id and creating a separate index for retry_id

src/iris/api.py
```diff
@@ -3311,6 +3311,32 @@ def on_post(self, req, resp):
...
+                           FROM `message` WHERE `id` = :msg_id
+                        ''',
+                        {'msg_id': msg_id}
+                    ).lastrowid
+                    session.execute(
+                        '''INSERT INTO `twilio_retry` (`message_id`, `retry_id`)
+                           VALUES (:msg_id, :retry_id)
+                        ''',
+                        {'msg_id': msg_id, 'retry_id': retry_id})
+            session.commit()
```
jrgp: I think this session.commit() should be inside the `if not is_retry` block as that one has the INSERTs, and we have a session.commit() earlier in the function

test/e2etest.py
```diff
@@ -2636,6 +2637,52 @@ def test_twilio_delivery_update(fake_message_id):
...
+    re = requests.post(base_url + 'twilio/deliveryupdate', data={'MessageSid': message_sid, 'MessageStatus': 'failed'})
+    assert re.status_code == 204
+
+    with iris_ctl.db_from_config(sample_db_config) as (conn, cursor):
+        cursor.execute('SELECT `message_id`, `retry_id` FROM `twilio_retry` WHERE `message_id` = %s', fake_message_id)
+        assert cursor.rowcount == 1
+        row = cursor.fetchone()
+        assert row[0] == fake_message_id
+        retry_id = row[1]
+        cursor.execute('SELECT * FROM `message` WHERE `id` = %s', retry_id)
```
jrgp: can we avoid `select *` ? if you just need to check for presence using .rowcount, you can do something like `select 1` or `select null`

db/schema_0.sql
dwang159: I think two unique indexes are the constraints we want. Each message can have exactly one retry entry (enforced by message_id PK), and no message can be a retry for more than one message (unique retry_id). Composite key would let us have multiple retries, which might make sense in the future, but keeping unique retry_id is still a good thing to have.

jrgp: That makes sense

PR #260, opened 2017-08-23
Utilize a separate history table for (un)claiming incidents

jrgp:
- Avoid UPDATE'ing the `active` and `owner_id` columns of the `incident`
  table by the API, to avoid deadlocks

- Instead, insert either "claim" or "unclaim" actions into a history
  table

- Show the claim/unclaim actions for a given incident on the incident
  view page

- Add/update various tests for the various pieces that these changes touch

### DIFF
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -150,12 +150,12 @@ incident_columns = {
     'id': '`incident`.`id` as `id`',
     'plan': '`plan`.`name` as `plan`',
     'plan_id': '`incident`.`plan_id` as `plan_id`',
-    'active': '`incident`.`active` as `active`',
+    'active': 'LEAST(`incident`.`active`, IFNULL(`incident_claim_action`.`name`, "unclaim") != "claim") as `active`',
     'updated': 'UNIX_TIMESTAMP(`incident`.`updated`) as `updated`',
     'application': '`application`.`name` as `application`',
     'context': '`incident`.`context` as `context`',
     'created': 'UNIX_TIMESTAMP(`incident`.`created`) as `created`',
-    'owner': '`target`.`name` as `owner`',
+    'owner': 'IF(`incident_claim_action`.`name` = "claim", `target`.`name`, NULL) as `owner`',
     'current_step': '`incident`.`current_step` as `current_step`',
 }
 
@@ -163,7 +163,7 @@ incident_filters = {
     'id': '`incident`.`id`',
     'plan': '`plan`.`name`',
     'plan_id': '`incident`.`plan_id`',
-    'active': '`incident`.`active`',
+    'active': 'LEAST(`incident`.`active`, IFNULL(`incident_claim_action`.`name`, "unclaim") != "claim")',
     'updated': '`incident`.`updated`',
     'application': '`application`.`name`',
     'context': '`incident`.`context`',
@@ -181,23 +181,30 @@ incident_filter_types = {
 }
 
 incident_query = '''SELECT %s FROM `incident`
+LEFT JOIN `incident_claim` ON `incident_claim`.`incident_id` = `incident`.`id` AND `incident_claim`.`id` =
+          (SELECT `id` FROM `incident_claim` WHERE `incident_claim`.`incident_id` = `incident`.`id` ORDER BY `time` DESC, `id` DESC LIMIT 1)
+LEFT JOIN `incident_claim_action` ON `incident_claim_action`.`id` = `incident_claim`.`action_id`
+LEFT JOIN `target` ON `target`.`id` = `incident_claim`.`target_id`
 JOIN `plan` ON `incident`.`plan_id` = `plan`.`id`
-LEFT OUTER JOIN `target` ON `incident`.`owner_id` = `target`.`id`
 JOIN `application` ON `incident`.`application_id` = `application`.`id`'''
 
-single_incident_query = '''SELECT `incident`.`id` as `id`,
+single_incident_query = '''
+SELECT `incident`.`id` as `id`,
     `incident`.`plan_id` as `plan_id`,
     `plan`.`name` as `plan`,
     UNIX_TIMESTAMP(`incident`.`created`) as `created`,
     UNIX_TIMESTAMP(`incident`.`updated`) as `updated`,
     `incident`.`context` as `context`,
-    `target`.`name` as `owner`,
     `application`.`name` as `application`,
     `incident`.`current_step` as `current_step`,
-    `incident`.`active` as `active`
+    LEAST(`incident`.`active`, IFNULL(`incident_claim_action`.`name`, "unclaim") != "claim") as `active`,
+    IF(`incident_claim_action`.`name` = "claim", `target`.`name`, NULL) as `owner`
 FROM `incident`
+LEFT JOIN `incident_claim` ON `incident_claim`.`incident_id` = `incident`.`id` AND `incident_claim`.`id` =
+          (SELECT `id` FROM `incident_claim` WHERE `incident_claim`.`incident_id` = `incident`.`id` ORDER BY `time` DESC, `id` DESC LIMIT 1)
+LEFT JOIN `incident_claim_action` ON `incident_claim_action`.`id` = `incident_claim`.`action_id`
+LEFT JOIN `target` ON `target`.`id` = `incident_claim`.`target_id`
 JOIN `plan` ON `incident`.`plan_id` = `plan`.`id`
-LEFT OUTER JOIN `target` ON `incident`.`owner_id` = `target`.`id`
 JOIN `application` ON `incident`.`application_id` = `application`.`id`
 WHERE `incident`.`id` = %s'''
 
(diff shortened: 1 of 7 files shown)


src/iris/api.py
```diff
@@ -181,23 +181,30 @@ def ts_to_sql_datetime(ts):
 }
 
 incident_query = '''SELECT %s FROM `incident`
+LEFT JOIN `incident_claim` ON `incident_claim`.`incident_id` = `incident`.`id` AND `incident_claim`.`id` =
+          (SELECT `id` FROM `incident_claim` WHERE `incident_claim`.`incident_id` = `incident`.`id` ORDER BY `time` DESC, `id` DESC LIMIT 1)
```
dwang159: EXPLAIN in MySQL is still giving me DEPENDENT SUBQUERY on lines like this. It's likely that when the incident claim table grows larger, these queries will slow down. How about something like:
```
LEFT JOIN `incident_claim` ON `incident_claim`.`incident_id` = `incident`.`id`
LEFT JOIN (SELECT `id`, MAX(`time`), `incident_id` FROM `incident_claim` GROUP BY `incident_id`) `t` ON `t`.`incident_id` = `incident`.`id`
```
Although this requires our MySQL to disable ONLY_FULL_GROUP_BY (which I believe is already the case).

src/iris/api.py
```diff
@@ -181,23 +181,30 @@ def ts_to_sql_datetime(ts):
...
     `incident`.`context` as `context`,
-    `target`.`name` as `owner`,
     `application`.`name` as `application`,
     `incident`.`current_step` as `current_step`,
-    `incident`.`active` as `active`
+    LEAST(`incident`.`active`, IFNULL(`incident_claim_action`.`name`, "unclaim") != "claim") as `active`,
+    IF(`incident_claim_action`.`name` = "claim", `target`.`name`, NULL) as `owner`
 FROM `incident`
+LEFT JOIN `incident_claim` ON `incident_claim`.`incident_id` = `incident`.`id` AND `incident_claim`.`id` =
+          (SELECT `id` FROM `incident_claim` WHERE `incident_claim`.`incident_id` = `incident`.`id` ORDER BY `time` DESC, `id` DESC LIMIT 1)
```
dwang159: Correlated subquery (same as above)

src/iris/api.py
```diff
@@ -3473,28 +3493,40 @@ def on_get(self, req, resp):
             'total_messages_sent_today': 'SELECT COUNT(*) FROM `message` WHERE `sent` >= CURDATE()',
             'total_active_users': 'SELECT COUNT(*) FROM `target` WHERE `type_id` = (SELECT `id` FROM `target_type` WHERE `name` = "user") AND `active` = TRUE',
             'pct_incidents_claimed_last_month': '''SELECT ROUND(
-                                                   (SELECT COUNT(*) FROM `incident`
+                                                   (SELECT COUNT(*)
+                                                    FROM `incident`
+                                                    LEFT JOIN `incident_claim` ON `incident_claim`.`incident_id` = `incident`.`id` AND `incident_claim`.`id` =
+                                                              (SELECT `id` FROM `incident_claim` WHERE `incident_claim`.`incident_id` = `incident`.`id` ORDER BY `time` DESC, `id` DESC LIMIT 1)
```
dwang159: Dependent subquery

src/iris/api.py
```diff
@@ -3473,28 +3493,40 @@ def on_get(self, req, resp):
...
-                                                    AND `active` = FALSE
+                                                    AND LEAST(`incident`.`active`, IFNULL(`incident_claim_action`.`name`, "unclaim") != "claim") = FALSE
                                                     AND NOT isnull(`owner_id`)) /
                                                    (SELECT COUNT(*) FROM `incident`
                                                     WHERE `created` > (CURRENT_DATE - INTERVAL 29 DAY)
                                                     AND `created` < (CURRENT_DATE - INTERVAL 1 DAY)) * 100, 2)''',
             'median_seconds_to_claim_last_month': '''SELECT @incident_count := (SELECT count(*)
                                                                                 FROM `incident`
+                                                                                LEFT JOIN `incident_claim` ON `incident_claim`.`incident_id` = `incident`.`id` AND `incident_claim`.`id` =
+                                                                                          (SELECT `id` FROM `incident_claim`
```
dwang159: Dependent subquery

src/iris/api.py
```diff
@@ -3473,28 +3493,40 @@ def on_get(self, req, resp):
...
                                                                                 AND NOT ISNULL(`owner_id`)
                                                                                 AND NOT ISNULL(`updated`)),
                                                             @row_id := 0,
                                                             (SELECT CEIL(AVG(time_to_claim)) as median
                                                             FROM (SELECT `updated` - `created` as time_to_claim
                                                                   FROM `incident`
+                                                                  LEFT JOIN `incident_claim` ON `incident_claim`.`incident_id` = `incident`.`id` AND `incident_claim`.`id` =
+                                                                            (SELECT `id` FROM `incident_claim`
+                                                                             WHERE `incident_claim`.`incident_id` = `incident`.`id` ORDER BY `time` DESC, `id` DESC LIMIT 1)
```
dwang159: Dependent subquery

src/iris/api.py
```diff
@@ -3548,19 +3580,27 @@ def on_get(self, req, resp, app_name):
...
+                                                                                AND LEAST(`incident`.`active`, IFNULL(`incident_claim_action`.`name`, "unclaim") != "claim") = FALSE
                                                                                 AND NOT ISNULL(`owner_id`)
                                                                                 AND NOT ISNULL(`updated`)
                                                                                 AND `application_id` = %(application_id)s),
                                                             @row_id := 0,
                                                             (SELECT CEIL(AVG(time_to_claim)) as median
                                                             FROM (SELECT `updated` - `created` as time_to_claim
                                                                   FROM `incident`
+                                                                  LEFT JOIN `incident_claim` ON `incident_claim`.`incident_id` = `incident`.`id` AND `incident_claim`.`id` =
+                                                                            (SELECT `id` FROM `incident_claim`
```
dwang159: Dependent subquery

src/iris/utils.py
```diff
@@ -76,9 +76,14 @@ def parse_response(response, mode, source):
                                                             JOIN `target` on `target`.`id` = `message`.`target_id`
                                                             JOIN `incident` on `incident`.`id` = `message`.`incident_id`
                                                             JOIN `target_type` on `target`.`type_id` = `target_type`.`id`
+                                                            LEFT JOIN `incident_claim` ON `incident_claim`.`incident_id` = `incident`.`id` AND `incident_claim`.`id` =
+                                                                      (SELECT `id` FROM `incident_claim`
```
dwang159: Dependent subquery

src/iris/utils.py
```diff
@@ -197,26 +213,43 @@ def claim_incident(incident_id, owner, session=None):
...
+
     claimed = set()
     not_claimed = set()
 
-    for incident_id, active in session.execute('SELECT `id`, `active` FROM `incident` WHERE `id` IN :incident_ids', {'incident_ids': incident_ids}):
-        if active == 1:
-            not_claimed.add(incident_id)
-        else:
+    check_claimed_sql = '''
+      SELECT `id`, (
```
dwang159: Could be made more efficient with joins, but if :incident_ids is small, it won't really matter since the dataset will be so small.

dwang159:
My comments might be MySQL version dependent if they've made their query optimizer smarter recently. My desktop is on 5.5.51.

src/iris/api.py
jrgp: The problem I ran into with using MAX(`time`) is that it is quite common for there to be several incident_claim actions in the same second. Because of that, I think ordering on `id` is safer.

dwang159: You can add another join to get MAX(id) per incident_id/time, but I'm not sure if that ends up being worth it. Maybe we try it this way and switch to joins if performance becomes an issue. Or if @fellyns could weigh in?

Alternately, since we're already using id as a tiebreaker, maybe we just get MAX(id) per incident_id, although that seems a little hacky

src/iris/utils.py
jrgp: Updated this to use join instead of subquery.

src/iris/api.py
fellyns: Given how we shard mysql the ordinality can't really be relied on. But using it as a "tie breaker" for time should be ok - fuzzy, but ok - as long as we are deterministic it will be fine.

dwang159:
Let's give this a shot at some point and see how the perf is. Is there an easy way to rollback? Seems like the migration may be a little tricky. Maybe we could do some load testing in staging?

jrgp:
I'm really hesitant about this, because:

- there's no way this logic and implementation will be as fast or performant as what we have now; the only gain is auditing and a less likely occurrence of deadlocks

- we removed most of the deadlocks and associated 5xxs by adding retry's and catching them when they happen

- rollback will be messy because, by design, this code isn't altering the `active` column

jrgp:
This PR and its potential performance and reliability caveats scare me, so I'm closing for now.

PR #261, opened 2017-08-25
Raise 400 on invalid claim owner

dwang159:


src/iris/api.py
```diff
@@ -1383,6 +1412,16 @@ def on_post(self, req, resp, incident_id):
             raise HTTPBadRequest('"owner" field required')
 
         with db.guarded_session() as session:
+            owner_valid = \
```
jrgp: This is a little tricky because we allow users to "unclaim" by passing NULL as the "owner" key. Can you only do the check if the owner isn't NULL/None, and add the NULL functionality to the route comment doc?

Also, can you avoid the trailing backslash to terminate the line?

src/iris/api.py
```diff
@@ -1383,6 +1412,16 @@ def on_post(self, req, resp, incident_id):
...
+                session.execute('''SELECT EXISTS(
+                                       SELECT 1
+                                       FROM `target`
+                                       WHERE `target`.`name` = :owner
+                                       AND `type_id` = (SELECT `id` FROM `target_type` WHERE `name` = 'user'))''',
+                                 {'owner': owner}).scalar()
+            if not owner_valid:
+                session.close()
+                raise HTTPBadRequest('Invalid claim: no matching owner')
```
jrgp: Also: can you add a e2etest for this?

test/e2etest.py
```diff
@@ -1396,6 +1402,9 @@ def test_post_incident_without_apps(sample_user, sample_team, sample_template_na
 
     # The application in sample_application_name2 does not have any sample_template_name templates, so this
     # will fail
+    print sample_application_name2
```
jrgp: do we need these prints here?

test/e2etest.py
```diff
@@ -1237,6 +1237,12 @@ def test_post_incident(sample_user, sample_team, sample_application_name, sample
     assert re.status_code == 200
     assert re.json() == {'owner': sample_user, 'incident_id': incident_id, 'active': False}
 
+    # Invalid claim owner
+    re = requests.post(base_url + 'incidents/%d' % (incident_id, ), json={
```
jrgp: Should be able to do:

    base_url + 'incidents/%d' % incident_id

Instead of:

    base_url + 'incidents/%d' % (incident_id, )

test/e2etest.py
```diff
@@ -1237,6 +1237,12 @@ def test_post_incident(sample_user, sample_team, sample_application_name, sample
...
+        'owner': 'nonexistent-user-foobar',
```
jrgp: At top of the file, we have a `invalid_user` variable set. Might be best to use that for consistency.

test/e2etest.py
```diff
@@ -1237,6 +1237,12 @@ def test_post_incident(sample_user, sample_team, sample_application_name, sample
...
+    }, headers={'Authorization': 'hmac %s:abc' % sample_application_name})
+    assert re.status_code == 400
```
jrgp: Can you check that we're getting back the error message `Invalid claim: no matching owner` in addition to the 400? This way we can validate against other types of 4xxs.

src/iris/api.py
```diff
@@ -1375,13 +1375,57 @@ def on_get(self, req, resp, incident_id):
...
+            connection = db.engine.raw_connection()
+            cursor = connection.cursor()
+            cursor.execute(
+                '''SELECT EXISTS(
+                     SELECT 1
+                     FROM `target`
+                     WHERE `target`.`name` = %s
+                     AND `type_id` = (SELECT `id` FROM `target_type` WHERE `name` = 'user'))''',
+                owner)
+            owner_valid = cursor.fetchone()[0]
```
jrgp: can you add a `cursor.close()` here? I'll merge after that.

jrgp: Bump :)

jrgp: Bump :)

dwang159: Oops

src/iris/api.py
```diff
@@ -1387,6 +1416,22 @@ def on_post(self, req, resp, incident_id):
...
 
+        if owner is not None:
+            connection = db.engine.raw_connection()
+            cursor = connection.cursor()
+            cursor.execute(
+                '''SELECT EXISTS(
+                     SELECT 1
+                     FROM `target`
+                     WHERE `target`.`name` = %s
+                     AND `type_id` = (SELECT `id` FROM `target_type` WHERE `name` = 'user'))''',
```
jrgp: Sorry, one more thing I just noticed: this will allow inactive users. Should we add another WHERE checking the `target`.`active` column?

jrgp: bump

PR #267, opened 2017-09-12
Support ldap login with start tls

ccl0326:
as title.

### DIFF
--- a/src/iris/ui/auth/ldap.py
+++ b/src/iris/ui/auth/ldap.py
@@ -10,16 +10,21 @@ class Authenticator:
     def __init__(self, config):
         root = os.path.abspath('./')
         self.ldap_url = config['auth']['ldap_url']
-        self.cert_path = os.path.join(root, config['auth']['ldap_cert_path'])
+        self.start_tls = config['auth']['start_tls']
+        if not self.start_tls:
+            self.cert_path = os.path.join(root, config['auth']['ldap_cert_path'])
         self.user_suffix = config['auth']['ldap_user_suffix']
         self.authenticate = self.ldap_auth
         if config.get('debug'):
             self.authenticate = self.debug_auth
 
     def ldap_auth(self, username, password):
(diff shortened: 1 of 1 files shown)


src/iris/ui/auth/ldap.py
```diff
@@ -10,16 +10,21 @@ class Authenticator:
     def __init__(self, config):
         root = os.path.abspath('./')
         self.ldap_url = config['auth']['ldap_url']
-        self.cert_path = os.path.join(root, config['auth']['ldap_cert_path'])
+        self.start_tls = config['auth']['start_tls']
```
jrgp: can you update the sample config to add the new usage of this?

src/iris/ui/auth/ldap.py
```diff
@@ -10,16 +10,21 @@ class Authenticator:
...
+        if not self.start_tls:
+            self.cert_path = os.path.join(root, config['auth']['ldap_cert_path'])
         self.user_suffix = config['auth']['ldap_user_suffix']
         self.authenticate = self.ldap_auth
         if config.get('debug'):
             self.authenticate = self.debug_auth
 
     def ldap_auth(self, username, password):
-        ldap.set_option(ldap.OPT_X_TLS_CACERTFILE, self.cert_path)
+        if not self.start_tls:
```
jrgp: is start tls mutually exclusive with passing a cacert file?

src/iris/ui/auth/ldap.py
dwang159: Switch to config['auth'].get('start_tls') to maintain existing behavior without need for config change

jrgp: ^ agreed :)

baryluk:
I can pick this up and address the remaining issues. I agree, it should use `.get()` so it works with existing config.

I also would like an option to ignore server certs / name. I have a prototype code to do that.

PR #268, opened 2017-09-12
Use ldap_user_search_string instead of ldap_user_suffix

ccl0326:
Use ldap_user_suffix for ldap login is not make sense. Use format string for ldap DN is better.

### DIFF
--- a/src/iris/ui/auth/ldap.py
+++ b/src/iris/ui/auth/ldap.py
@@ -11,7 +11,7 @@ class Authenticator:
         root = os.path.abspath('./')
         self.ldap_url = config['auth']['ldap_url']
         self.cert_path = os.path.join(root, config['auth']['ldap_cert_path'])
-        self.user_suffix = config['auth']['ldap_user_suffix']
+        self.user_search_string = config['auth']['ldap_user_search_string']
         self.authenticate = self.ldap_auth
         if config.get('debug'):
             self.authenticate = self.debug_auth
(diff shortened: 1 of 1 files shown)


src/iris/ui/auth/ldap.py
```diff
@@ -11,7 +11,7 @@ def __init__(self, config):
         root = os.path.abspath('./')
         self.ldap_url = config['auth']['ldap_url']
         self.cert_path = os.path.join(root, config['auth']['ldap_cert_path'])
-        self.user_suffix = config['auth']['ldap_user_suffix']
+        self.user_search_string = config['auth']['ldap_user_search_string']
```
dwang159: Can you alter the config.yaml example files to account for this change?

ccl0326: @dwang159 There is no `config.yam` example in project, and the `config.dev.yaml` does not  have any ldap login config in it.

dwang159: Hm, you're right. I was thinking about the linkedin/oncall repo, sorry.

jrgp: Can you make this support both config options? (So we and other people can avoid changing our configs)

Eg if you only specify `ldap_user_suffix`, default `self.user_search_string` to `'%%s%s' % config['auth']['ldap_user_suffix']` to preserve the old behavior

dwang159:
Looks like this is implemented by #341, closing here.

PR #276, opened 2017-10-04
Add stats per application

dwang159:
Get message counts per app, per mode. Also get call retry count

### DIFF
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -3598,6 +3598,11 @@ class ApplicationStats(object):
                                                                   ORDER BY time_to_claim) as time_to_claim
                                                             WHERE (SELECT @row_id := @row_id + 1)
                                                             BETWEEN @incident_count/2.0 AND @incident_count/2.0 + 1)''',
+            'total_call_retry_last_month': '''SELECT COUNT(*)
+                                              FROM `twilio_retry` JOIN `message` ON `message`.`id` = `twilio_retry`.`message_id`
+                                              WHERE `sent` > (CURRENT_DATE - INTERVAL 29 DAY)
+                                              AND `sent` < (CURRENT_DATE - INTERVAL 1 DAY)
+                                              AND `application_id` = %(application_id)s'''
         }
 
         query_data = {'application_id': app['id']}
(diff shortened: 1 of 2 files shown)


src/iris/api.py
```diff
@@ -3598,6 +3598,11 @@ def on_get(self, req, resp, app_name):
                                                                   ORDER BY time_to_claim) as time_to_claim
                                                             WHERE (SELECT @row_id := @row_id + 1)
                                                             BETWEEN @incident_count/2.0 AND @incident_count/2.0 + 1)''',
+            'total_call_retry_last_month': '''SELECT COUNT(*)
```
jrgp: Can you add this key to the `test_app_stats()` e2e test?

src/iris/api.py
```diff
@@ -3670,6 +3675,18 @@ def on_get(self, req, resp, app_name):
...
+        for mode in cache.modes:
+            cursor.execute('''SELECT COUNT(*) FROM `message`
+                              WHERE `sent` > (CURRENT_DATE - INTERVAL 29 DAY)
+                              AND `sent` < (CURRENT_DATE - INTERVAL 1 DAY)
+                              AND `application_id` = %(application_id)s
+                              AND `mode_id` = (SELECT `id` FROM `mode` WHERE `name` = %(mode)s)''',
+                           {'application_id': app['id'], 'mode': mode})
+            num_sent = cursor.fetchone()
+            if num_sent is not None:
+                num_sent = num_sent[-1]
```
jrgp: 1) why `-1` instead of `0`?
2) If no rows are returned, the stats[] entry will be set to None, which might cause problems with things expecting to get ints back for these keys. Should probably set it to 0 if we get back None from the cursor

dwang159: 1. -1 is consistent with the earlier stats query, but also it looks like it does that specifically for the median_seconds_to_claim to work. 0 it is.
2. Count(*) should always return 1 row, so I'm not sure if None ever happens. Even if it does happen, I think None would be better than 0, since 0 implies no messages were sent and None implies something crazy happened.

PR #278, opened 2017-10-06
Add mobile authentication/device registration

dwang159:
Add QR code in user settings page for enabled users to get a mobile key. Also add device registration endpoint for storing device IDs (used for push notifications).

### DIFF
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -14,6 +14,7 @@ import re
 import os
 import datetime
 import logging
+import pyqrcode
 import jinja2
 from jinja2.sandbox import SandboxedEnvironment
 from urlparse import parse_qs
@@ -25,6 +26,7 @@ import falcon.uri
 
 from collections import defaultdict
 from streql import equals
+from base64 import b64encode
 
 from . import db
 from . import utils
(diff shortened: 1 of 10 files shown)


src/iris/ui/static/js/iris.js
```diff
@@ -1762,6 +1777,27 @@ iris = {
...
+    },
+    regenMobileKey: function () {
+      $.ajax({
+        url: this.data.mobileKeyUrl + this.data.user,
+        method: 'POST',
+        contentType: 'application/json',
+        dataType: 'html'
+      }).done(function(data){
+        iris.createAlert('Mobile key regenerated.', 'success');
+        $('#mobile-key-container').html('<img width="300" height="300" src="' + data + '">');
```
jrgp: fix `data` as above

src/iris/ui/static/js/iris.js
```diff
@@ -1762,6 +1777,27 @@ iris = {
       });
       this.redrawApplicationDropdown();
     },
+    createMobile: function() {
+      $.ajax({
+        url: this.data.mobileKeyUrl + this.data.user,
+        method: 'GET',
+        dataType: 'html'
+      }).done(function(data){
+        $('#mobile-key-container').html('<img width="300" height="300" src="' + data + '">')
```
jrgp: this is vulnerable to XSS in case `data` gets compromised. you should create the image object and programmatically assign the src value. eg:

```
$('#mobile-key-container').empty(); // clear it first if needed
$('#mobile-key-container').append($('<img width="300" height="300">').attr('src', data));
```

it's possible that @shoxv is aware of a less verbose way of doing this, but the gist is that html() is extremely dangerous, especially when interpolating data directly. It's pretty much HTML's equivalent to SQL injection

src/iris/api.py
```diff
@@ -3646,6 +3648,125 @@ def on_get(self, req, resp, app_name):
...
+            conn.commit()
+        except:
+            raise falcon.HTTPError("DB error")
+        finally:
+            conn.close()
+
+        return self.png_to_datauri(image_as_str)
+
+    def on_get(self, req, resp, user):
+        time.sleep(4)
```
jrgp: can we avoid the sleep? we don't want to risk deliberately exhausting our workers.

src/iris/api.py
```diff
@@ -3646,6 +3648,125 @@ def on_get(self, req, resp, app_name):
...
+        conn = db.engine.raw_connection()
+        cursor = conn.cursor()
+
+        cursor.execute(
+            '''SELECT `mobile_key` FROM `user`
+               WHERE `target_id` = (SELECT `id` FROM `target` WHERE `name` = %s
+                 AND `type_id` = (SELECT `id` FROM `target_type` WHERE `name` = 'user'))''',
+            user)
+        data = {'secret': cursor.fetchone()[0],
```
jrgp: this will raise an exception if fetchone() doesn't return anything. should handle that.

src/iris/api.py
```diff
@@ -3646,6 +3648,125 @@ def on_get(self, req, resp, app_name):
...
+            cursor.execute(
+                '''UPDATE `user` SET `mobile_key` = %s
+                   WHERE `target_id` = (SELECT `id` FROM `target` WHERE `name` = %s
+                     AND `type_id` = (SELECT `id` FROM `target_type` WHERE `name` = 'user'))''',
+                (secret, user))
+            if cursor.rowcount == 0:
+                raise HTTPNotFound()
+            # Commit your changes in the database
+            conn.commit()
+        except:
```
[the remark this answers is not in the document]
dwang159: Fixed in line below

src/iris/api.py
```diff
@@ -3694,6 +3696,130 @@ def on_get(self, req, resp, app_name):
...
+        cursor = conn.cursor()
+
+        cursor.execute(
+            '''SELECT `mobile_key` FROM `user`
+               WHERE `target_id` = (SELECT `id` FROM `target` WHERE `name` = %s
+                 AND `type_id` = (SELECT `id` FROM `target_type` WHERE `name` = 'user'))''',
+            user)
+        secret = cursor.fetchone()
+        if secret is None:
+            raise HTTPNotFound()
```
jrgp: i don't think conn is being closed in case we raise? should probably close conn and its cursor right after we fetchone() as it isn't used in the rest of the method.

src/iris/api.py
```diff
@@ -3694,6 +3696,138 @@ def on_get(self, req, resp, app_name):
...
+        secret = cursor.fetchone()
+        if secret is None:
+            cursor.close()
+            conn.close()
+            raise HTTPNotFound()
+        else:
+            secret = secret[0]
+        data = {'secret': secret,
+                'api_url': self.mobile_url}
+        cursor.close()
```
jrgp: You can move the cursor and connection closing calls to right after cursor.fetchone() as it doesn't look like they're used below, that way you can avoid closing them in two places.

PR #285, opened 2017-10-23
Make sender use individual queues per message mode

jrgp:
This way, if all emails back up we won't affect delivery, or marking
messages as sent, for other modes such as call/sms. This will result
in no more duplicate sms/calls when emails get backed up.

- Use one queue per message mode, rather than one global send_queue
- Disable some sender unit tests as this is a slightly large refactor and
  those tests are very dependent on how the sender works

### DIFF
--- a/test/test_sender.py
+++ b/test/test_sender.py
@@ -8,6 +8,7 @@
 import time
 from iris.bin.sender import init_sender
 import msgpack
+import gevent.queue
 
 
 def test_configure(mocker):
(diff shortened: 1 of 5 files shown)


test/test_sender.py
```diff
@@ -103,6 +104,7 @@ def init_queue_with_item(queue, item=None):
         queue.put(item)
 
 
+@pytest.mark.skip('re-enable when we rewrite these tests')
```
dwang159: Can we redo the tests before merging the RB? Otherwise I don't think they'll ever get rewritten...

jrgp: Fixed this

PR #302, opened 2017-11-03
Dynamically use MX records for each SMTP connection pooling

jrgp:
- Optionally periodically scale up/down the number of smtp tasks per mx record according
  to DNS, relying on some new config parameters attached to the smtp vendor.

- Reuse SMTP connection per smtp worker task

- Upon shutdown, try to cleanly stop all workers after they have finished sending their last
  message.

- Add vendor cleanup method, in this case used for iris_smtp to cleanly disconnect from the SMTP
  server when it is time to die

### DIFF
--- a/src/iris/bin/sender.py
+++ b/src/iris/bin/sender.py
@@ -12,10 +12,11 @@ import socket
 import gevent
 import signal
 import setproctitle
+import copy
 
 from collections import defaultdict
 from iris.plugins import init_plugins
-from iris.vendors import IrisVendorManager
+from iris.vendors import IrisVendorManager, iris_smtp
 from iris.sender import auditlog
 from iris import metrics
 from uuid import uuid4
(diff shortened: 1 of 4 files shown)


src/iris/bin/sender.py
```diff
@@ -1164,12 +1327,34 @@ def mock_gwatch_renewer():
...
+        shutdown_started = True
+        logger.info('Shutting server..')
 
     # Immediately release all locks and give up any master status and slave presence
     if coordinator:
         coordinator.leave_cluster()
 
+    # Try to wait for all sender greenlets to die. Get a list, instead of generator,
+    # of all task objects so we can iterate over it multiple times.
+    all_worker_tasks = list(itertools.chain.from_iterable(
```
fellyns: Not sure what the goal of this line is - would just list concatenation not suffice here?

jrgp: The values themselves are lists. Using itertools to flatten them.

PR #319, opened 2017-11-21
sort the incident context

wleese:
While integrating alertmanager, I noticed that having many context fields results in a very messy view of the incident. Sort to be more pleasing on the eye.

### DIFF
--- a/src/iris/ui/static/js/iris.js
+++ b/src/iris/ui/static/js/iris.js
@@ -1368,6 +1368,15 @@ iris = {
     init: function(){
       var location = window.location.pathname.split('/'),
           path = this.data.id = location[location.length - 1];
+
+      Handlebars.registerHelper('eachSorted', function(context, options) {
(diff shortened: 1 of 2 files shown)


fellyns:
Thanks for the PR @wleese!

I'll flag this with the rest of the team to see what they think, but I wanted to mention why we don't already have something like this.

When you integrate a new application with Iris there is a place to add templates for how the context should be rendered in the UI. The unordered dumping of the context is meant to be more of a debugging view until you get those templates written. That said there's little harm in sorting it (maybe slightly larger JS payload and extra JS processing to sort).

wleese:
Ah, I get it now. Guess i’ll be adding some templates then. Couldn’t find anything about this in the docs - I thought templating was only done for alert messages.

wleese:
I'll close this, as I personally have no need for it anymore now that I've written proper templates.

PR #320, opened 2017-11-22
Add (Prometheus) Alertmanager support

wleese:
This is done directly in Iris in order to minimize dependencies for such a crucial function.

In Alertmanager, one would simply use the alert label "iris_plan" to select the Iris plan.

HMAC authentication is disabled, as this isn't supported by Alertmanager.
Instead, application name and key are passed in the query string.

An example snippet that can be added to Alertmanager's config:

```
        receivers:
        - name: 'iris-team1'
          webhook_configs:
            - url: http://iris:16649/v0/alertmanager?application=test-app&key=sdffdssdf
```

Related to issue https://github.com/linkedin/iris/issues/110


fellyns:
Thanks for the PR @wleese!

I really like the functionality here, but I don't know how much we should embed it directly in the core code. Ideally something like this would be pluggable/optional for people that are not using prometheus (like us as we don't use Prometheus inside LinkedIn).

The common approach to integrate Alertmanager webhooks with an existing service seems to be to have a sidecar service that handles the translation of Alertmanager messages to whatever is used by the service. Simple though inelegant and adds additional complexity.

I think the best approach here would be to extract this out into something configurable. Perhaps a new type of plugin for generic webhooks that we load at runtime based on config (like a mapping of route to handler).

Also, I do have an Iris Golang client that we use internally that I can opensource if you want to pursue the sidecar method and use Go as Prometheus does.

wleese:
I’ll take a shot at making this plugable - I prefer not to take the sidecar approach. If Iris is ever to be a pagerduty replacement, it’d have to be convenient to use. Separation in multiple components has many benefits, but ease of use shouldn’t have to suffer.

wleese:
test still failing despite https://github.com/linkedin/iris/commit/e5237db963d49759be3ba5b5f57d7821e603688b

dwang159:
Is e5237db in your commit history? You may need to merge master back into your branch again.

wleese:
Silly me, thanks :)

dwang159:
Looks like flake8 is unhappy, otherwise LGTM.

src/iris/api.py
```diff
@@ -684,7 +685,16 @@ def debug_auth(self, req, resp, resource, params):
 
         # Proceed with authenticating this route as a third party application
         try:
-            app, client_digest = req.get_header('AUTHORIZATION', '')[5:].split(':', 1)
+            # Ignore HMAC requirements for custom webhooks
+            if len(req.env['PATH_INFO'].split("/")) > 2 and req.env['PATH_INFO'].split("/")[2] == 'webhooks':
```
jrgp: Can we do `if req.env['PATH_INFO'].startswith('/v0/webhooks/')` instead? Should be more efficient than splitting twice and it's less code.

src/iris/api.py
```diff
@@ -700,6 +710,28 @@ def process_resource(self, req, resp, resource, params):  # pragma: no cover
         if resource.allow_read_no_auth and method == 'GET':
             return
 
+        # Ignore HMAC requirements for custom webhooks
+        if len(req.env['PATH_INFO'].split("/")) > 2 and req.env['PATH_INFO'].split("/")[2] == 'webhook':
```
jrgp: Can we do `if req.env['PATH_INFO'].startswith('/v0/webhooks/')` instead? Should be more efficient than splitting twice and it's less code.

src/iris/api.py
```diff
@@ -684,7 +685,16 @@ def debug_auth(self, req, resp, resource, params):
 
         # Proceed with authenticating this route as a third party application
         try:
-            app, client_digest = req.get_header('AUTHORIZATION', '')[5:].split(':', 1)
+            # Ignore HMAC requirements for custom webhooks
+            if len(req.env['PATH_INFO'].split("/")) > 2 and req.env['PATH_INFO'].split("/")[2] == 'webhooks':
+                qs = parse_qs(req.env['QUERY_STRING'])
```
jrgp: Instead of using parse_qs/etc, use falcon's built-in helpers if possible:

`app = req.get_param('application', required=True)`

src/iris/api.py
```diff
@@ -700,6 +710,28 @@ def process_resource(self, req, resp, resource, params):  # pragma: no cover
         if resource.allow_read_no_auth and method == 'GET':
             return
 
+        # Ignore HMAC requirements for custom webhooks
+        if len(req.env['PATH_INFO'].split("/")) > 2 and req.env['PATH_INFO'].split("/")[2] == 'webhook':
+            qs = parse_qs(req.env['QUERY_STRING'])
+
+            app_name = qs['application'][0]
```
jrgp: Instead of using parse_qs/etc and manually decoding the query string, please use falcon's built-in helpers if possible.

```
app = req.get_param('application', required=True)
api_key = req.get_param('key', required=True)
```

test/test_webhook_alertmanager.py
```diff
@@ -0,0 +1,140 @@
...
+        "commonAnnotations": {
+            "description": "testcommonannotation",
+            "summary": "testcommonsummary"
+            },
+        "externalURL": "http://a9fabe732a60:9093",
+        "version": "4",
+        "groupKey": "{}:{alertname=\"testalert\"}"
+    }
+    am_webhook.validate_post(fake_post)
+
```
jrgp: to satisfy flake8 we need two empty lines here rather than 1

src/iris/webhooks/alertmanager.py
```diff
@@ -0,0 +1,101 @@
...
+
+from iris import db
+
+logger = logging.getLogger(__name__)
+
+
+class alertmanager(object):
+    allow_read_no_auth = False
+
+    def on_get(self, req, resp):
```
jrgp: If you leave on_get undefined, falcon will reply back with a HTTP 405 which is probably good enough

src/iris/api.py
```diff
@@ -700,6 +708,26 @@ def process_resource(self, req, resp, resource, params):  # pragma: no cover
...
+            app = cache.applications.get(app_name)
+            if not app:
+                logger.warn('Tried authenticating with nonexistent app: "%s"', app_name)
+                raise HTTPUnauthorized('Authentication failure', '', [])
+
+            req.context['app'] = app
+
+            # determine if we're correctly using an application key
+            api_key = req.get_param('key', required=True)
+            if not api_key:
```
jrgp: When `required=True`, Falcon will raise a 4xx if the param is not defined. You should be able to remove the `if not api_key` check

src/iris/api.py
```diff
@@ -684,7 +685,14 @@ def debug_auth(self, req, resp, resource, params):
 
         # Proceed with authenticating this route as a third party application
         try:
-            app, client_digest = req.get_header('AUTHORIZATION', '')[5:].split(':', 1)
+            # Ignore HMAC requirements for custom webhooks
+            if req.env['PATH_INFO'].startswith('/v0/webhooks/'):
+                app = req.get_param('application', required=True)
+                if not app:
```
jrgp: When `required=True`, Falcon will raise a 4xx if the param is not provided. You should be able to remove the `if not app` check

src/iris/api.py
```diff
@@ -700,6 +708,26 @@ def process_resource(self, req, resp, resource, params):  # pragma: no cover
...
+                raise HTTPUnauthorized('Authentication failure', '', [])
+
+            req.context['app'] = app
+
+            # determine if we're correctly using an application key
+            api_key = req.get_param('key', required=True)
+            if not api_key:
+                logger.warn('Did not provide application key for "%s"', app_name)
+                raise HTTPUnauthorized('Authentication failure', '', [])
+            if not api_key == str(app['key']):
```
jrgp: should probably be `if not equals(api_key, str(app['key']))` to prevent memory timing attacks. (`equals` is already imported and used elsewhere in the auth checking)

src/iris/api.py
```diff
@@ -700,6 +706,19 @@ def process_resource(self, req, resp, resource, params):  # pragma: no cover
         if resource.allow_read_no_auth and method == 'GET':
             return
 
+        # Ignore HMAC requirements for custom webhooks
+        if req.env['PATH_INFO'].startswith('/v0/webhooks/'):
+            app_name = req.get_param('application', required=True)
+            app = cache.applications.get(app_name)
```
jrgp: really sorry to ask for another change but the `if not app` check should still be present here to make sure the one we did get from the query string is present in the applications dict.

jrgp:
There we go. After the test I'll merge.

wleese:
That's for putting up with my incremental patch work - all the feedback was more than appreciated and'll hopefully result in better PRs moving forward.

jrgp:
No worries whatsoever. Thanks for keeping up with my suggestions :)

houqp:
Thanks @wleese for the patch, making it so much easier to integrate with our prometheus cluster 👍

PR #322, opened 2017-11-29
Configurable user timezones for UI dates

jrgp:
- Add moment-timezone with tz data from oncall
- Start using the unused user_settings table


src/iris/api.py
```diff
@@ -3048,6 +3079,66 @@ def on_get(self, req, resp, username):
...
+            raise HTTPBadRequest('Invalid json in post body')
+
+        # Add more keys here if we eventually need more settings
+        allowed_keys = ('timezone', )
+
+        connection = db.engine.raw_connection()
+        cursor = connection.cursor()
+
+        for name, value in data.iteritems():
+            if name not in allowed_keys:
```
fellyns: With the "name == 'timezone'" logic below we are essentially checking if this is the string timezone twice on every iteration. Not the end of the world, but there are more efficient ways to validate the settings.

src/iris/api.py
```diff
@@ -806,6 +822,21 @@ def process_admin_acl(self, req, resource, params):
...
+            return
+
+        connection = db.engine.raw_connection()
+        cursor = connection.cursor()
+        cursor.execute(get_username_settings_query, req.context['username'])
+        settings = dict(cursor)
+        cursor.close()
+        connection.close()
+
+        req.context['user_settings'].update(settings)
```
fellyns: Why update an empty dict vs setting it directly?

PR #324, opened 2017-12-01
hipchat support

wleese:
Messages will be posted to the provided room id.
Users will be 'mentioned', enabling them to receive their alerts.

https://github.com/linkedin/iris/issues/321

Known issue: currently the iris username is used as the hipchat username when sending alerts. I'll make this configurable once I understand how the hipchat username field can be stored in Oncall.


src/iris/vendors/iris_hipchat.py
```diff
@@ -0,0 +1,72 @@
...
+        self.token = self.config.get('auth_token')
+        self.room_id = self.config.get('room_id')
+        self.debug = self.config.get('debug')
+        self.endpoint_url = self.config.get('base_url')
+
+        self.params = {'auth_token': self.token}
+        self.notification_url = '{0}/v2/room/{1}/notification'.format(self.endpoint_url, self.room_id)
+        self.headers = {
+            'Content-type': 'application/json',
+            }
```
jrgp: slight nitpick: can you put the closing brace on the same level of indention as self.headers?

src/iris/vendors/iris_hipchat.py
```diff
@@ -0,0 +1,72 @@
...
+            'color': 'red',
+            'notify': 'true',
+            'message_format': "text",
+        }
+        return self.message_dict
+
+    def send_message(self, message):
+        start = time.time()
+        payload = ujson.dumps(self.get_message_payload(message))
+        logger.info('tmpdebug msg: %s', ujson.dumps(message))
```
jrgp: should remove or move this logger.info to the `if self.debug` section, or take it out of the PR and just leave it uncommitted for local testing (what we do in this situation)

src/iris/vendors/iris_hipchat.py
```diff
@@ -0,0 +1,72 @@
...
+    def get_message_payload(self, message):
+        """Send notification to specified HipChat room"""
+        clean_message = "@{0} {1}".format(message['destination'], message['body'])
+        self.message_dict = {
+            'message': clean_message,
+            'color': 'red',
+            'notify': 'true',
+            'message_format': "text",
+        }
+        return self.message_dict
```
jrgp: probably should avoid setting message_dict as a property of the class, and just return it.

that way the function becomes pure and doesn't modify external state (makes testing cleaner and more reproducible) and won't ever cause problems if we decide to call the same vendor object more than once concurrently

src/iris/vendors/iris_hipchat.py
```diff
@@ -0,0 +1,72 @@
...
+        payload = ujson.dumps(self.get_message_payload(message))
+        logger.info('tmpdebug msg: %s', ujson.dumps(message))
+        if self.debug:
+            logger.info('debug: %s', self.message_dict)
+        else:
+            try:
+                response = requests.post(self.notification_url,
+                                         headers=self.headers,
+                                         params=self.params,
+                                         data=payload,
```
jrgp: you can avoid manually making the json string and passing it to `data=` and instead just do `json=payload` and it will encode in json for you

test/test_iris_vendor_hipchat.py
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env python
+# -*- coding:utf-8 -*-
+
+from iris.vendors.iris_hipchat import iris_hipchat
+import ujson as json
```
jrgp: this json import probably can be removed as it isn't used anywhere in this file

src/iris/vendors/iris_hipchat.py
```diff
@@ -0,0 +1,70 @@
...
+            'notify': 'true',
+            'message_format': "text",
+        }
+        return message_dict
+
+    def send_message(self, message):
+        start = time.time()
+        payload = self.get_message_payload(message)
+        if self.debug:
+            logger.info('debug: %s', self.message_dict)
```
jrgp: I think `self.message_dict` here should be `payload`?

src/iris/vendors/iris_hipchat.py
```diff
@@ -0,0 +1,70 @@
...
+import logging
+import requests
+import time
+from iris.constants import IM_SUPPORT
+
+logger = logging.getLogger(__name__)
+
+
+class iris_hipchat(object):
+    supports = frozenset(['hipchat'])
```
jrgp: Can we add a `HIPCHAT_SUPPORT = 'hipchat'` to constants.py and use that here and in the iris.constants import above?

wleese:
All comments addressed and tests fixed.

jrgp:
Thanks!

jrgp:
Whoops meant to squash rather than rebase

PR #329, opened 2017-12-06
Add commands to fill in the gaps

wleese:
The output directory needs to be created manually, but was unmentioned.
Also, use tail to remove the invalid json caused by gen_packer_cfg.py outputting an informative message


jrgp:
Thanks!

PR #333, opened 2017-12-07
Add documentation on Oncall syncing

wleese:
and add alertmanager documentation

### DIFF
--- a/docs/source/admin_manual.rst
+++ b/docs/source/admin_manual.rst
@@ -16,6 +16,7 @@ Service Integration
 -------------------
 .. toctree::
    nagios
+   alertmanager
 
 
 Iris components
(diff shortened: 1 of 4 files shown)


adityavijjanT:
I am getting below error while running sync on iris.

File "/iris/src/iris/bin/sync_targets.py", line 145, in <dictcomp>
    if user['active']}
TypeError: string indices must be integers
Makefile:13: recipe for target 'targets' failed
make: *** [targets] Error 1

Below are the logs from oncall server.

[10/Jan/2019:04:24:05 +0000] "GET /api/v0/users?fields=name&fields=contacts&fields=active HTTP/1.1" 401 85 "-" "python-requests/2.20.0"

PR #335, opened 2017-12-09
Reset metrics only updated by master senders

anirajk:
Sometimes, alerting systems depend on metrics from the master node.
It's better to not emit those metrics or emit a zero value if the current node is not master


src/iris/bin/sender.py
```diff
@@ -1477,6 +1477,16 @@ def init_sender(config):
                                   slaves=config['sender'].get('slaves', []))
 
 
+def reset_master_only_metrics():
```
jrgp: I think it would be enough to add these to the `default_sender_metrics` dict at the top of the file, as all of those get set to 0 with each iteration by metrics.emit. Looks like `new_incidents_cnt` is already there.

PR #336, opened 2017-12-11
annotated config for config.dev.yaml

wleese:
Alternatively to https://github.com/linkedin/iris/pull/334, annotate the existing config.dev.yaml instead of introducing another file.


dwang159:
I think I also prefer the single-file approach. LGTM

PR #339, opened 2017-12-12
Catch Template Syntax exception while getting plans

anirajk:
We should continue in the event of bad templating syntax for one application within a plan


src/iris/sender/cache.py
```diff
@@ -219,13 +219,17 @@ def __getitem__(self, key):
...
+                        try:
+                            tracking_template[application] = {
+                                'email_subject': self.template_env.from_string(application_templates['email_subject']),
+                                'email_text': self.template_env.from_string(application_templates['email_text']),
+                            }
+                            html_template = application_templates.get('email_html')
+                            if html_template:
+                                tracking_template[application]['email_html'] = self.template_env.from_string(html_template)
+                        except jinja2.exceptions.TemplateSyntaxError:
+                            logger.info('[-] error parsing Plan template for %s: %s', key, application)
```
jrgp: Can you make this a logger.exception instead?

PR #345, opened 2017-12-30
Include iris ui static files in source distribution

martinwangjian:
Include iris ui template and static files when `python setup.py sdist`.
To fix error `jinja2.exceptions.TemplateNotFound`


jrgp:
This should be good. Let's see if the test passes on the latest master.

jrgp:
👍

PR #350, opened 2018-01-10
Clean inactive incidents from queue

dwang159:
Seems like our current method is somewhat leaky. Change to align
messages queue with database status

### DIFF
--- a/src/iris/bin/sender.py
+++ b/src/iris/bin/sender.py
@@ -507,15 +507,7 @@ def aggregate(now):
             cursor.close()
             connection.close()
 
-            inactive_message_ids = aggregated_message_ids - active_message_ids
             l = len(active_message_ids)
-            logger.info('[x] dropped %s messages from claimed incidents, %s remain for %r',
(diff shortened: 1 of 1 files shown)


src/iris/bin/sender.py
```diff
@@ -535,6 +527,19 @@ def aggregate(now):
...
+    connection = db.engine.raw_connection()
+    cursor = connection.cursor()
+    cursor.execute('SELECT `id` FROM `message` WHERE active=1')
+    active_message_ids = {r[0] for r in cursor}
+    cursor.close()
+    connection.close()
+    inactive = 0
+    for message_id in messages:
+        if message_id not in active_message_ids:
+            del messages[message_id]
```
jrgp: probably should do `messages.pop(message_id, None)` to avoid a KeyError in case it isn't present in that dict

dwang159:
Closing, instead I'll look into cleaning up `messages` in `mark_message_as_sent`.

PR #354, opened 2018-01-16
Flexible tracking types

wleese:
Allow all contact modes to be used for Incident Tracking.

https://github.com/linkedin/iris/issues/353

Still a bit rough around the edges, with lot's of the email specific stuff not to mention UI.
Posting early for feedback.

### DIFF
--- a/src/iris/bin/sender.py
+++ b/src/iris/bin/sender.py
@@ -451,9 +451,15 @@ def escalate():
                 tracking_message = {
                     'noreply': True,
                     'destination': tracking_key,
-                    'mode': tracking_type,
-                    'body': tracking_message['body']
+                    'mode': tracking_type
                 }
+                try:
+                    body = app_tracking_template['body'].render(**context)
+                except Exception as e:
+                    body = 'plan %s - tracking notification body failed to render: %s' % (plan['name'], str(e))
+                    logger.exception(body)
+                tracking_message['body'] = body
+
             message_send_enqueue(tracking_message)
     cursor.close()
 
--- a/src/iris/bin/sender.py
+++ b/src/iris/bin/sender.py
@@ -418,36 +418,38 @@ def escalate():
                 'plan_id': plan_id,
                 'application': application,
             }
-            if tracking_type == 'email':
-                tracking_message = {
-                    'noreply': True,
-                    'destination': tracking_key,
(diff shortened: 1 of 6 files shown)


wleese:
Slowly progressing. Javascript painful.

wleese:
Ready for review.

src/iris/bin/sender.py
```diff
@@ -447,7 +447,20 @@ def escalate():
...
+            else:
+                tracking_message = {
+                    'noreply': True,
+                    'destination': tracking_key,
+                    'mode': tracking_type
+                }
+                try:
+                    body = app_tracking_template['body'].render(**context)
+                except Exception as e:
+                    body = 'plan %s - tracking notification body failed to render: %s' % (plan['name'], str(e))
```
jrgp: You can remove the explicit str() as %s will do that for you

src/iris/ui/templates/plan.html
```diff
@@ -255,6 +258,13 @@ <h4><i class="icon-plus icon">+</i> Add application tracking template</h4>
           <div class="notification-header">HTML</div>
           <pre class="notification-body">{{breakLines email_html}}</pre>
         </div>
+        {{/isEqual}}
+        {{^isEqual ../tracking_type 'email'}}
```
jrgp: I think this should be a # not ^ ?

wleese: While I cannot point to documentation - I couldn't find any. I found a comment on stackoverflow (?) mentioning this usage negates the helpers behavior. I've tested this and it works as intended as far as I can tell.
Also see: https://github.com/linkedin/oncall/pull/128#discussion_r160308027

wleese:
@dwang159 cannot reproduce.

- Go to Plans
- Create New Plan
- Expand Incident Tracking
(At that point, the following settings are prefilled: Type: "call" and  Target: (grey)example@example.com)
- Type in a target
- Click "Add Application Tracking Template"
- Pick an application, fill out the template and Save.
- Go back to Plans tab, open your new plan. .. all tracking settings should be visible (likewise right after the save when the plan is displayed in read mode).

dwang159:
My fault, my database was incorrectly configured. Testing again, seems to work fine.

PR #361, opened 2018-01-23
Allow disabling gwatch

wleese:
- Allow disabling gwatch by introducing config['sender']['disable_gwatch_renewer_task']. False if left unconfigured to retain behavior in existing setups
- Rename config['skipgmailwatch'] to config['sender']['mock_gwatch_renewer_task'], to better reflect its behavior
- Document both variables and set "disable_gwatch_renewer_task: True" so newcomers get a better default

https://github.com/linkedin/iris/issues/360

### DIFF
--- a/src/iris/bin/sender.py
+++ b/src/iris/bin/sender.py
@@ -1504,7 +1504,8 @@ def main():
 
     maintain_workers(config)
 
-    gwatch_renewer_task = config['sender'].get('disable_gwatch_renewer_task', False)
+    disable_gwatch_renewer = config['sender'].get('disable_gwatch_renewer_task', False)
+    gwatch_renewer_task = None
     prune_audit_logs_task = None
 
     interval = 60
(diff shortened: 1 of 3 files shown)


src/iris/bin/sender.py
```diff
@@ -1504,7 +1504,7 @@ def main():
 
     maintain_workers(config)
 
-    gwatch_renewer_task = None
+    gwatch_renewer_task = config['sender'].get('disable_gwatch_renewer_task', False)
```
dwang159: Now that this takes values other than None, we should refactor line 1531 to `if not disable_gwatch_renewer and gwatch_renewer_task is not None`. Setting this to a boolean before initializing it to a task is somewhat confusing on first glance, especially since "True" results in no gwatch renewer task.

dwang159:
On second thought, renaming/moving `skipgmailwatch` breaks backwards compatibility and doesn't seem like a particularly pressing issue. Can we look up `skipgmailwatch` as a default value if `disable_gwatch_renewer_task` doesn't exist? We can look into deprecating/moving this config value in the future, with more notice.

dwang159:
Agreed, I think skipsend and skipgmailwatch should live under the sender config, so we'll look into doing a cleanup at some point.

PR #362, opened 2018-01-23
allow notification to room and add hipchat metrics

wleese:
Originally the hipchat vendor only supported sending notifications to a single room with a "mention" of the user who it's meant for.

This PR allows another scenarios: if the destination value has the correct syntax, we send to the specified room_id. If a username was mentioned (@ prefix) we also do the mention.

Also fixes missing hipchat metrics.

### DIFF
--- a/src/iris/vendors/iris_hipchat.py
+++ b/src/iris/vendors/iris_hipchat.py
@@ -28,15 +28,13 @@ class iris_hipchat(object):
         self.debug = self.config.get('debug')
         self.endpoint_url = self.config.get('base_url')
 
-        self.params = {'auth_token': self.token}
-        self.notification_url = '{0}/v2/room/{1}/notification'.format(self.endpoint_url, self.room_id)
         self.headers = {
             'Content-type': 'application/json',
         }
 
-    def get_message_payload(self, message):
+    def get_message_payload(self, message, mention):
         """Send notification to specified HipChat room"""
-        clean_message = "@{0} {1}".format(message['destination'], message['body'])
+        clean_message = "{0} {1}".format(mention, message['body'])
         message_dict = {
             'message': clean_message,
             'color': 'red',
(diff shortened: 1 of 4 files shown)


src/iris/vendors/iris_hipchat.py
```diff
@@ -45,16 +43,58 @@ def get_message_payload(self, message):
         }
         return message_dict
 
+    def parse_destination(self, destination):
+        """Determine room_id, token and user to mention
+           We accept 3 formats:
+             - Just a mention (@testuser)
+             - Room_id and Token (12341:a20asdfgjahdASDfaskw)
```
dwang159: Should this `:` be a `;` in this line and the next?

src/iris/vendors/iris_hipchat.py
```diff
@@ -45,16 +43,58 @@ def get_message_payload(self, message):
...
+        if destination.startswith("@"):
+            mention = destination
+        elif ";" in destination:
+            dparts = destination.split(";")
+            if len(dparts) == 3:
+                try:
+                    int(dparts[0])
+                    room_id = dparts[0]
+                except ValueError:
+                    pass
```
dwang159: Should this be `pass`? I'm not sure what a non-integer room id would imply. I'd suggest logging a warning that an invalid destination was passed and that the vendor is falling back to the default room.

src/iris/vendors/iris_hipchat.py
```diff
@@ -45,16 +43,56 @@ def get_message_payload(self, message):
...
+            elif len(dparts) == 2:
+                try:
+                    room_id = int(dparts[0])
+                except ValueError:
+                    logger.error("Invalid destination: %s. Using default room_id", destination)
+                token = dparts[1]
+            else:
+                logger.error("Invalid destination: %s. Contains %s fields.", destination, len(dparts))
+        else:
+                logger.error("Invalid destination: %s. Neither @user or ; separated value.", destination)
```
jrgp: Is this logger.error indented four spaces too far ?

PR #365, opened 2018-01-24
messagebird support (sms only)

wleese:
SMS only messagebird support.

https://github.com/linkedin/iris/issues/364

### DIFF
--- a/src/iris/vendors/iris_messagebird.py
+++ b/src/iris/vendors/iris_messagebird.py
@@ -57,14 +57,13 @@ class iris_messagebird(object):
             try:
                 response = requests.post(self.endpoint_url,
                                          headers=self.headers,
-                                         data=json.dumps(payload),
+                                         json=payload,
                                          proxies=self.proxy)
                 if response.status_code == 201:
                     return time.time() - start
                 else:
-                    logger.error('Failed to send message to messagebird: %d',
-                                 response.status_code)
-                    logger.error("Response: %s", response.content)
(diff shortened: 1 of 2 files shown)


src/iris/vendors/iris_messagebird.py
```diff
@@ -0,0 +1,72 @@
...
+        start = time.time()
+        payload = self.get_message_payload(message)
+
+        if self.debug:
+            logger.info('debug: %s', payload)
+        else:
+            try:
+                response = requests.post(self.endpoint_url,
+                                         headers=self.headers,
+                                         data=json.dumps(payload),
```
jrgp: You could instead do:

    json=payload

And requests will encode to json for you

src/iris/vendors/iris_messagebird.py
```diff
@@ -0,0 +1,72 @@
...
+                                         proxies=self.proxy)
+                if response.status_code == 201:
+                    return time.time() - start
+                else:
+                    logger.error('Failed to send message to messagebird: %d',
+                                 response.status_code)
```
jrgp: Can you please consolidate these two loggers into a single one ?

jrgp:
Cool, looks like you need to rip out the `import json` at the top and then it'll be good.

jrgp:
Coolio. Thanks for the patch! Now we have some variety in case we want to move off of twilio

wleese:
No problem - thanks for the patience.

PR #366, opened 2018-01-24
Indicate in UI why we failed to expand too large ldap lists

jrgp:
- Introduce a custom exception for passing deliberate failures from role lookup
  modules to sender, and use that for passing across the cause

### DIFF
--- a/src/iris/bin/sender.py
+++ b/src/iris/bin/sender.py
@@ -29,6 +29,7 @@ from iris.sender.message import update_message_mode
 from iris.sender.oneclick import oneclick_email_markup, generate_oneclick_url
 from iris import cache as api_cache
 from iris.sender.quota import ApplicationQuota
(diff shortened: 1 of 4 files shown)


src/iris/bin/sender.py
```diff
@@ -324,7 +332,8 @@ def create_messages(incident_id, plan_notification_id):
                      incident_id, plan_notification_id, role, target, names, name, priority_id)
 
         body = ('You are receiving this as you created this plan and we can\'t resolve'
-                ' %s of %s at this time.\n\n') % (role, target)
+                ' %s of %s at this time%s.\n\n') % (role, target, ': %s' % lookup_fail_reason if lookup_fail_reason else '')
```
dwang159: Formatting will be off here. Put the third %s after the `.`

PR #369, opened 2018-01-26
Clean messages queue on send

dwang159:
Before sending, ensure messages have been removed from the `messages` cache/queue.


jrgp:
Will be interesting to see what kind of impact this has

PR #372, opened 2018-02-02
Device registration, push notifications

dwang159:


src/iris/api.py
```diff
@@ -18,7 +18,9 @@
 from jinja2.sandbox import SandboxedEnvironment
 from urlparse import parse_qs
 import ujson
-from falcon import HTTP_200, HTTP_201, HTTP_204, HTTPBadRequest, HTTPNotFound, HTTPUnauthorized, HTTPForbidden, HTTPFound, HTTPInternalServerError, API
+from falcon import (before, HTTP_200, HTTP_201, HTTP_204, HTTPBadRequest,
```
jrgp: I think it'd be cleaner to use `falcon.before` instead of using a global ambiguous `before`

src/iris/api.py
```diff
@@ -3999,6 +4001,44 @@ def on_get(self, req, resp, app_name):
...
+        platform = data.get('platform')
+
+        if user is None or registration_id is None or os is None:
+            raise HTTPBadRequest('Missing parameters for adding device')
+
+        # Open database connection
+        conn = db.engine.raw_connection()
+        cursor = conn.cursor()
+
+        cursor.execute('''INSERT IGNORE INTO device (registration_id, user_id, platform)
```
jrgp: Should do error handling here to clean up the DB connection and log exception during failures

db/schema_0.sql
```diff
@@ -724,6 +724,17 @@ CREATE TABLE `mailing_list_membership` (
...
 ) ENGINE=InnoDB DEFAULT CHARSET=latin1;
 
+
+DROP TABLE IF EXISTS `device`;
+CREATE TABLE `device` (
+  `registration_id` VARCHAR(255) NOT NULL,
+  `user_id` BIGINT(20) NOT NULL,
+  `platform` VARCHAR(255) NOT NULL,
+  PRIMARY KEY (`registration_id`),
+  CONSTRAINT `device_user_id_ibfk` FOREIGN KEY (`user_id`) REFERENCES `user` (`target_id`)
```
jrgp: sorry I didn't mention this earlier: can we make these cascading deletes?

PR #375, opened 2018-02-15
Add endpoints to check valid target, target membership

jayesh92:
- Valid user endpoint caters to the need of checking the given
    target(user/mailing-list/team) is valid or not.
- Memebership endpoint takes a user and lists and returns whether the given
   user is a member of any of those.

Testing: make test.

### DIFF
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -3170,6 +3170,57 @@ class User(object):
         resp.body = ujson.dumps(user_data)
 
 
+class ValidTarget(object):
+    allow_read_no_auth = False
+    enforce_user = False
+
+    def on_get(self, req, resp, target_type):
+        valid_target_query = '''SELECT EXISTS(
+                                    SELECT 1
+                                    FROM `target`
+                                    WHERE `target`.`name` = %s)'''
+
+        try:
+            resp.body = ujson.dumps({'exists':
+                                     bool(db.engine.execute(valid_target_query,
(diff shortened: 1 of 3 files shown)


src/iris/api.py
```diff
@@ -3170,6 +3170,57 @@ def on_get(self, req, resp, username):
         resp.body = ujson.dumps(user_data)
 
 
+class ValidTarget(object):
+    allow_read_no_auth = True
```
jrgp: We want to set `allow_read_no_auth` to `False`, to make these two routes only work if you're using an API key or a logged in user.

src/iris/api.py
```diff
@@ -3170,6 +3170,57 @@ def on_get(self, req, resp, username):
...
+                                          WHERE `target_type`.`name` = 'mailing-list'
+                                          AND `_target`.`name` IN %s)))'''
+
+        try:
+            resp.body = ujson.dumps({'is_member':
+                                     bool(db.engine.execute(membership_query,
+                                                            (username,
+                                                             lists)).scalar())})
+        except Exception:
+            logger.exception('Error checking lists membership for target: %s, lists: %s' % (username, lists))
```
jrgp: You can do this instead:

    logger.exception('Error checking lists membership for target: %s, lists: %s', username, lists)

As the logger functions automatically use all extra arguments to format the string.

jayesh92:
@jrgp Updated Comments.

PR #378, opened 2018-02-26
Make message_id constraint on `response` cascade ops

jrgp:
Needed for pruning of incidents/messages. Missed this one when retention
functionality was added in #367


anirajk:
LGTM. might be worth looking at other places that might be missing cascaded updates.

PR #383, opened 2018-03-13
Bump webassets

jrgp:
- Old version has problems where it doesn't always generate asset URLs


dwang159:
Tests failed? Not sure why

jrgp:
Rebuilding without cache. I think the old venv having the previous version broke it.

PR #387, opened 2018-03-20
Add support for generating/accepting secondary API key

dwang159:


src/iris/api.py
```diff
@@ -778,37 +778,39 @@ def process_resource(self, req, resp, resource, params):  # pragma: no cover
...
+                            req.context['app'] = app
+                            if username_header:
+                                req.context['username'] = username_header
+                                return
+                # No successful HMACs match, fail auth.
+                if username_header:
+                    logger.warn('HMAC doesn\'t validate for app %s (passing username %s)', app['name'], username_header)
+                else:
+                    logger.warn('HMAC doesn\'t validate for app %s', app['name'])
+                    raise HTTPUnauthorized('Authentication failure', '', [])
```
jrgp: I might be missing something, but if debug is turned off and no app key is correct but you do pass a username header, i'm not sure if the request will be blocked if neither for loops `return`

src/iris/api.py
```diff
@@ -2695,6 +2697,69 @@ def on_get(self, req, resp, app_name):
         resp.body = ujson.dumps({'key': key})
 
 
+class ApplicationSecondaryKey(ApplicationKey):
```
jrgp: Why should this inherit from the other key class?

src/iris/api.py
```diff
@@ -2695,6 +2697,69 @@ def on_get(self, req, resp, app_name):
...
+                       JOIN `application` on `application`.`id` = `application_owner`.`application_id`
+                       WHERE `target`.`name` = :username
+                       AND `application`.`name` = :app_name''',
+                    {'app_name': app_name, 'username': req.context['username']}).scalar()
+                if not has_permission:
+                    raise HTTPForbidden('You don\'t have permissions to re-key this app.')
+
+        data = {
+            'app_name': app_name,
+            'new_key': hashlib.sha256(os.urandom(32)).hexdigest()
```
jrgp: As we're now doing this twice, we should probably abstract the key generation into a separate function.

src/iris/ui/static/js/iris.js
```diff
@@ -2245,6 +2251,20 @@ iris = {
         iris.createAlert('Unable to get key: ' + result.responseJSON.title);
       });
     },
+    showSecondaryKey: function() {
+      var self = this;
+      $(self.data.showSecondaryKeyButton).prop('disabled', true);
+      $.get('/v0/applications/' + self.data.application + '/secondary').done(function(result) {
+        self.data.model.secondaryKey = result.key;
+        if (result.key ==  null) {
```
jrgp: extra space after equals sign

src/iris/ui/static/js/iris.js
```diff
@@ -2424,6 +2447,23 @@ iris = {
         $rekeyBtn.prop('disabled', false);
       });
     },
+    genSecondaryApplication: function() {
```
jrgp: Would probably be more clear if this had "key" in the name

PR #388, opened 2018-03-26
Allow iris api url to be configured in config file instead of default http://127.0.0.1:16649

seuf:
When using iris api behind an https reverse proxy (nginx for example) the `get_local` function try to fetch information from `http://127.0.0.1:16649` instead of `https://{{ host }}:16649`.
This pull request allow iris api-url configuration directly from config file.
I've also renamed the `get_local` method to `get_iris_api`.

### DIFF
--- a/configs/config.dev.yaml
+++ b/configs/config.dev.yaml
@@ -5,6 +5,7 @@ server:
   host: 0.0.0.0
   port: 16649
   disable_auth: True # allow logins by existing users, but without checking credentials
+  local_api_url: http://localhost:16649
 
 db: &db
   conn:
--- a/src/iris/ui/__init__.py
+++ b/src/iris/ui/__init__.py
@@ -98,8 +98,8 @@ def get_flash(req):
 # are only retrievable by going through the API itself. Forge a request locally and keep the
 # beaker cookies so the current user can be authenticated. This will go away once each page
 # that makes use of window.appData is converted to use ajax for those values instead.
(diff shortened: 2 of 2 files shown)


configs/config.dev.yaml
```diff
@@ -5,6 +5,7 @@ server:
   host: 0.0.0.0
   port: 16649
   disable_auth: True # allow logins by existing users, but without checking credentials
+  api-url: http://localhost:16649
```
jrgp: Can you rename this to `local_api_url` ?

src/iris/ui/__init__.py
```diff
@@ -338,7 +338,7 @@ def on_post(self, req, resp):
             resp.status = falcon.HTTP_400
             return
 
-        app_json = get_local(req, 'applications/%s' % application)
+        app_json = get_iris_api(req, 'applications/%s' % application)
```
jrgp: Can you rename the function to `get_local_api` ?

src/iris/ui/__init__.py
```diff
@@ -392,12 +392,14 @@ def on_post(self, req, resp):
 
 
 def init(config, app):
+    global api_url
```
jrgp: can you rename this variable to `local_api_url` ?

seuf:
renamed variables and functions to ` local_api_url`

jrgp:
Thanks

PR #389, opened 2018-04-03
Add application stats daemon

dwang159:


src/iris/bin/app_stats.py
```diff
@@ -0,0 +1,78 @@
...
+        logger.exception('Failed writing pid to %s', pidfile)
+
+stats_reset = {
+    'task_failure': 0,
+}
+
+
+def set_app_stats(app, stats, connection, cursor):
+    for stat, val in stats.iteritems():
+        cursor.execute('''REPLACE INTO `application_stats` (`application_id`, `statistic`, `value`, `timestamp`)
```
jrgp: Instead of doing `REPLACE INTO`, can you refactor this to use `INSERT INTO ... ON DUPLICATE KEY UPDATE ..` ?

src/iris/bin/app_stats.py
```diff
@@ -0,0 +1,78 @@
...
+
+def main():
+    config = load_config()
+    metrics.init(config, 'iris-application-stats', stats_reset)
+    retention_settings = config.get('app-stats', {})
+    run_interval = int(retention_settings['run_interval'])
+    spawn(metrics.emit_forever)
+
+    db.init(config)
+    while 1:
```
jrgp: Can you make this `while True` instead of `while 1` for consistency with other parts of the code?

src/iris/api.py
```diff
@@ -3959,11 +3959,14 @@ def on_get(self, req, resp):
...
-        if not app:
-            raise HTTPNotFound()
+    def __init__(self, config):
+        cfg = config.get('app-stats', {})
+        # Calculate stats in real time (True), or query offline stats
+        # generated by the app stats daemon (False)
+        self.real_time = cfg.get('real_time', True)
 
+    @staticmethod
+    def calculate_app_stats(app, connection, cursor, fields_filter=None):
```
jrgp: Can we move this calculate_app_stats function to somewhere outside the Falcon route class? It's kind of odd for a daemon script to call a method that is part of a route

PR #390, opened 2018-04-04
Fix indentation issue in auth check

dwang159:


jrgp:
Merged. That method is getting pretty hairy. Might be a good idea to break it into smaller methods at some point and add tests.

PR #391, opened 2018-04-05
Skip null values in per-app stats

dwang159:
Sometimes claim percentages, median claim time, etc. have null
values (for apps with no incidents)


jrgp:
Why do we need to store rows with empty values ? Seems like a bit of a waste

PR #393, opened 2018-04-23
Add timeouts to result3 calls in sync_targets

dwang159:
Fix cases where we hang on LDAP query

### DIFF
--- a/src/iris/bin/sync_targets.py
+++ b/src/iris/bin/sync_targets.py
@@ -22,6 +22,7 @@ from ldap.controls import SimplePagedResultsControl
 from ldap.filter import escape_filter_chars
 import time
 ldap_pagination_size = 1000
+ldap_timeout = None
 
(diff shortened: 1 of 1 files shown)


src/iris/bin/sync_targets.py
```diff
@@ -304,7 +304,7 @@ def get_ldap_lists(l, search_strings, parent_list=None):
                              serverctrls=[req_ctrl],
                              attrlist=(search_strings['list_cn_field'], search_strings['list_name_field']),
                              filterstr=filterstr)
-        rtype, rdata, rmsgid, serverctrls = l.result3(msgid, resp_ctrl_classes=known_ldap_resp_ctrls)
+        rtype, rdata, rmsgid, serverctrls = l.result3(msgid, timeout=20, resp_ctrl_classes=known_ldap_resp_ctrls)
```
jrgp: Can we make both instances of this 20 configurable somewhere or a constant at the top of the file?

PR #405, opened 2018-06-08
rackspace webhook support

peebs:
So I wanted to float this to get feedback on the approach for adding support for a rackspace webhooks. In rackspace one can configure a notification to fire out a webhook. The main issue is that you have no real control over the json payload it sends, only the URL parameters. 

Looking at the alertmanager plugin, it seemed like the `plan_id` was the only value really necessary receive from the json payload. I opted to make an optional URL parameter that the rackspace webhook implementation will check. 

Is my assumption about `plan_id` being the only important value correct and is this an approach that would be accepted?

WIP as I plan to do more testing and add docs if this approach looks ok.


src/iris/webhooks/rackspace.py
```diff
@@ -0,0 +1,94 @@
...
+import logging
+import ujson
+from falcon import HTTP_201, HTTPBadRequest, HTTPNotFound
+
+from iris import db
+
+logger = logging.getLogger(__name__)
+
+
+class rackspace(object):
```
dwang159: This class has a lot of duplicated code with the alertmanager webhook. Could these inherit from a common webhook object?

src/iris/api.py
```diff
@@ -732,13 +735,18 @@ def process_resource(self, req, resp, resource, params):  # pragma: no cover
...
             app_name = req.get_param('application', required=True)
+
             app = cache.applications.get(app_name)
             if not app:
                 raise HTTPUnauthorized('Authentication failure',
                                        'Application not found', [])
 
             req.context['app'] = app
 
+            # optionally used and validated based on webhook impl
```
dwang159: Instead of finding the plan in the AuthMiddleware, do so in the rackspace `on_post()` method. Falcon passes the same `req` object to `on_post()`, so the implementation should be the same, but it's cleaner to group this with the other rackspace code.

test/test_webhook_rackspace.py
```diff
@@ -0,0 +1,125 @@
...
+                }
+            ]
+        },
+        "entity": {
+            "id": "enTEST",
+            "label": "Test Entity",
+            "ip_addresses": {
+                "default": "8.8.8.8"
+            },
+            "metadata": null,
```
jrgp: Need to change all of the:

`null` -> `None`
`false` -> `False`
`true` -> `True`

To convert this vaid json to valid python

pbx0:
Thanks for the review and sorry for the delay. Please take another look. I addressed all comments and added the changes as additional commits. I can squash down the commits after review depending on what is preferred.

src/iris/webhooks/alertmanager.py
```diff
@@ -45,52 +32,5 @@ def on_post(self, req, resp):
...
-                data).lastrowid
-
-            session.commit()
-            session.close()
-
-        resp.status = HTTP_201
-        resp.set_header('Location', '/incidents/%s' % incident_id)
-        resp.body = ujson.dumps(incident_id)
+        plan = alert_params['groupLabels']['iris_plan']
+        super().on_post(self, req, resp, plan)
```
dwang159: Does this work for Python 2?

test/test_webhook_rackspace.py
```diff
@@ -0,0 +1,120 @@
+# Copyright (c) LinkedIn Corporation. All rights reserved. Licensed under the BSD-2 Clause license.
```
dwang159: Can we add an end-to-end test for rackspace as well? Something like `test_post_incident()` in `e2etest.py`. I think we're missing coverage on `rackspace.on_post()` and `alertmanager.on_post()` currently, which would explain the `super()` issue.

allwyn-pradip:
can this be merged.. it seems to be stuck for quite some time.

allwyn-pradip (approved):

diegocepedaw:
@allwyn-pradip merged

PR #409, opened 2018-06-20
Updated iris slack vendor to be compatible with mattermost

seuf:
This PR update de iris slack vendor plugin to be compatible with mattermost :
 - all the payload is json dumped not only the attachement
 - the returned value is checked in text or in json

Tested with slack and mattermost (the claim button don't work in mattermost but that's okay for now).

### DIFF
--- a/src/iris/vendors/iris_slack.py
+++ b/src/iris/vendors/iris_slack.py
@@ -68,7 +68,7 @@ class iris_slack(object):
                 }
             ]
         }
-        return ujson.dumps([att_json])
+        return [att_json]
 
     def get_message_payload(self, message):
         slack_message = {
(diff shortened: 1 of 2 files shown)


src/iris/vendors/iris_slack.py
```diff
@@ -95,17 +95,27 @@ def send_message(self, message):
...
-                             data['error'])
+                try:
+                    data = response.json()
+                    if data['ok']:
+                        return time.time() - start
+                    # If message is invalid:
+                    #   {u'ok': False, u'error': u'invalid_arg_name'}
+                    logger.error('Received an error from slack api: %s',
+                                 data['error'])
+                except Exception:
```
dwang159: It looks like Mattermost's API is largely compatible with Slack's, but I don't think this is a good way to handle the incompatibilities. Can you make a new vendor instead? You can duplicate some of this code, but adapt it to better suit Mattermost (get rid of the buttons and handle the different response)

PR #418, opened 2018-07-13
Jinja validation for incident tracking

diegocepedaw:
added jinja validation to incident tracking body

### DIFF
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -614,7 +614,7 @@ def is_valid_tracking_settings(t, k, tpl):
     environment = SandboxedEnvironment()
     for app in tpl:
         try:
-            environment.from_string(tpl[app]['body'])          
+            environment.from_string(tpl[app]['body'])
         except Exception as e:
                 return False, 'Invalid jinja syntax in incident tracking text dy html: %s' % e
     return True, None
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -593,19 +593,19 @@ def is_valid_tracking_settings(t, k, tpl):
 
             try:
                 environment.from_string(tpl[app]['email_subject'])
-            except Exception as e:
+            except jinja2.TemplateSyntaxError as e:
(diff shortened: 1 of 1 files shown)


src/iris/api.py
```diff
@@ -610,6 +610,13 @@ def is_valid_tracking_settings(t, k, tpl):
     else:
         if t not in cache.modes:
             return False, 'Unknown tracking type: %s' % t
+
+    environment = SandboxedEnvironment()
+    for app in tpl:
+        try:
+            environment.from_string(tpl[app]['body'])          
+        except Exception as e:
```
dwang159: Catch `jinja2.TemplateSyntaxError`

src/iris/api.py
```diff
@@ -610,6 +610,13 @@ def is_valid_tracking_settings(t, k, tpl):
     else:
         if t not in cache.modes:
             return False, 'Unknown tracking type: %s' % t
+
+    environment = SandboxedEnvironment()
```
dwang159: This should go with the `else` in line 610, otherwise we validate email twice.

src/iris/api.py
dwang159: Probably update lines 601 and 608 too.

src/iris/api.py
```diff
@@ -610,6 +610,13 @@ def is_valid_tracking_settings(t, k, tpl):
...
+    for app in tpl:
+        try:
+            environment.from_string(tpl[app]['body'])          
+        except Exception as e:
+                return False, 'Invalid jinja syntax in incident tracking text dy html: %s' % e
```
dwang159: typo: "dy"

diegocepedaw: fixed

src/iris/api.py
diegocepedaw: fixed

src/iris/api.py
diegocepedaw: updated as well as 601 608 and 595

PR #420, opened 2018-07-17
Improve twilio delivery update reliability

dwang159:
Retry on deadlocks, move to raw_connection


src/iris/api.py
```diff
@@ -3782,42 +3782,58 @@ def on_post(self, req, resp):
...
+                try:
+                    affected = cursor.execute(
+                        '''UPDATE `twilio_delivery_status`
+                           SET `status` = %(status)s
+                           WHERE `twilio_sid` = %(sid)s''',
+                        {'sid': sid, 'status': status})
+                    connection.commit()
+                    break
+                except Exception:
+                    logger.exception('Failed running Twilio status query. (Try %s/%s)', i + 1, max_retries)
```
jrgp: When this fails do we need to create a new cursor object? I think I remember somewhere else in the codebase needing to create a new connection.cursor() whenever the old one fails and dies.

src/iris/api.py
```diff
@@ -3782,42 +3782,58 @@ def on_post(self, req, resp):
...
                         '''INSERT INTO `twilio_retry` (`message_id`, `retry_id`)
-                           VALUES (:msg_id, :retry_id)
+                           VALUES (%(msg_id)s, %(retry_id)s)
                         ''',
                         {'msg_id': msg_id, 'retry_id': retry_id})
-                    session.commit()
-            session.close()
+                    connection.commit()
+            cursor.close()
+        finally:
```
jrgp: Do we want to catch exceptions that might be raised and bubble up to falcon? It seems we only have the finally block.

src/iris/api.py
dwang159: It doesn't look like it. We use this pattern in iris-sender a few times, and we're never re-creating the cursor there.

PR #427, opened 2018-08-02
Change how inactive messages are cleaned, change log levels

dwang159:
Get all actives before aggregate task, clear all inactives at the
end. Saves a few DB queries, avoids leaks.
Also, demote non-fatal DB errors to warnings

### DIFF
--- a/src/iris/bin/sender.py
+++ b/src/iris/bin/sender.py
@@ -380,7 +380,7 @@ def deactivate():
     max_retries = 3
 
     # this deadlocks sometimes. try until it doesn't.
-    for i in xrange(max_retries):
+    for i in xrange(1, max_retries + 1):
         try:
             cursor.execute(GET_INACTIVE_IDS_SQL)
             ids = tuple(r[0] for r in cursor)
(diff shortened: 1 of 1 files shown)


src/iris/bin/sender.py
```diff
@@ -389,7 +389,10 @@ def deactivate():
                 connection.commit()
                 break
         except Exception:
-            logger.exception('Failed running deactivate query. (Try %s/%s)', i + 1, max_retries)
+            if i == max_retries - 1:
```
jrgp: instead of all `i + 1` and `i -1` logic, what if we did `xrange(1, max_retries + 1)` instead of needing to alter `i` every time we reference it?

src/iris/bin/sender.py
```diff
@@ -558,6 +553,14 @@ def aggregate(now):
                 logger.info('[-] purged %s from messages %s remaining', active_message_ids, len(messages))
             del queues[key]
             sent[key] = now
+    inactive_message_ids = set(messages.keys()) - all_actives
```
jrgp: I think either of these would work instead of casting the .keys() list to a set: `set(messages)` or `messages.viewkeys()`

PR #434, opened 2018-08-17
Precompute global iris stats

diegocepedaw:
Add the option to precompute global iris stats and retrieve them from the db instead of calculating them for every request.

### DIFF
--- a/db/schema_0.sql
+++ b/db/schema_0.sql
@@ -751,18 +751,10 @@ CREATE TABLE `application_stats` (
 
 DROP TABLE IF EXISTS `global_stats`;
 CREATE TABLE `global_stats` (
-  `id` BIGINT(20) NOT NULL AUTO_INCREMENT,
-  `median_seconds_to_claim_last_month` FLOAT,
-  `total_incidents_today` BIGINT(20),
-  `total_active_users` BIGINT(20),
-  `total_plans` BIGINT(20),
-  `total_messages_sent` BIGINT(20),
-  `total_applications` BIGINT(20),
-  `pct_incidents_claimed_last_month` FLOAT,
-  `total_incidents` BIGINT(20),
-  `total_messages_sent_today` BIGINT(20),
+  `statistic` VARCHAR(255) NOT NULL,
+  `value` FLOAT NOT NULL,
   `timestamp` DATETIME NOT NULL,
-  PRIMARY KEY (`id`)
+  PRIMARY KEY (`statistic`)
 ) ENGINE=InnoDB DEFAULT CHARSET=latin1;
 
 
--- a/db/schema_0.sql
+++ b/db/schema_0.sql
@@ -749,6 +749,22 @@ CREATE TABLE `application_stats` (
     ON DELETE CASCADE ON UPDATE CASCADE
 ) ENGINE=InnoDB DEFAULT CHARSET=latin1;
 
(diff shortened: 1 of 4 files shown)


src/iris/app_stats.py
```diff
@@ -133,3 +133,62 @@ def calculate_app_stats(app, connection, cursor, fields_filter=None):
         if count_stat not in stats:
             stats[count_stat] = 0
     return stats
+
+
+def calculate_gobal_stats(connection, cursor, fields_filter=None):
```
dwang159: global

db/schema_0.sql
```diff
@@ -749,6 +749,22 @@ CREATE TABLE `application_stats` (
     ON DELETE CASCADE ON UPDATE CASCADE
 ) ENGINE=InnoDB DEFAULT CHARSET=latin1;
 
+DROP TABLE IF EXISTS `global_stats`;
```
dwang159: model on `application_stats` instead, so we can add global stats without needing a schema change.

src/iris/api.py
```diff
@@ -4002,67 +4003,43 @@ def on_get(self, req, resp):
...
+                WHERE `timestamp` = (SELECT MAX(`timestamp`) from `global_stats` gs2 WHERE gs2.`timestamp` = gs1.`timestamp`)
+            '''
+            cursor.execute(query)
+
+            if cursor.rowcount == 0:
+                logger.error('Error retrieving global stats from db')
+                raise HTTPInternalServerError('Error retrieving global stats from db')
+
+            # build dictionary of stat names and values
+            desc = cursor.description
```
jrgp: If you do `cursor = connection.cursor(db.dict_cursor)`, the rows cursor.fetchall() returns will be a dict mapping column names to values, which alleviates having to construct that dict manually here.

But then, that likely breaks the rest of the uses of this cursor as those all expect to not be using a dict cursor.

As you already know the fields which will be returned, as you listed them explicitly in your query, you could avoid relying on this metadata and construct the dict manually. eg:

```
columns = ('median_seconds_to_claim_last_month', 'total_incidents_today', 'total_active_users', 'total_plans', 'total_messages_sent', 'total_applications', 'pct_incidents_claimed_last_month', 'total_incidents', 'total_messages_sent_today', 'timestamp')
row = cursor.fetchone()
stats = {column: row[i] for column, i in enumerate(columns)}
```

src/iris/api.py
```diff
@@ -4002,67 +4003,43 @@ def on_get(self, req, resp):
...
+        if self.real_time:
+            stats = app_stats.calculate_gobal_stats(connection, cursor, fields_filter=fields_filter)
+
+        else:
+            # select the most current global stats entry
+            query = '''
+                SELECT `median_seconds_to_claim_last_month`, `total_incidents_today`, `total_active_users`, `total_plans`, `total_messages_sent`,
+                `total_applications`, `pct_incidents_claimed_last_month`, `total_incidents`, `total_messages_sent_today`, `timestamp`
+                FROM `global_stats` gs1
+                WHERE `timestamp` = (SELECT MAX(`timestamp`) from `global_stats` gs2 WHERE gs2.`timestamp` = gs1.`timestamp`)
```
jrgp: you can avoid this subquery and where clause by sorting on the timestamp descending. eg  `ORDER BY `timestamp desc` DESC LIMIT 1`.

PR #436, opened 2018-08-27
Added password for demo user in README.md

k1-hedayati:


dwang159:
Technically any password will suffice with the dummy authentication module, but `demo` seems as good as any.

k1-hedayati:
Oh didn't test that :)

PR #437, opened 2018-08-27
Show amount of time a step will take in the UI

dwang159:
Somewhat unintuitive, since it's the minimum wait time of the listed
notifications (not the max)

### DIFF
--- a/src/iris/ui/static/js/iris.js
+++ b/src/iris/ui/static/js/iris.js
@@ -154,6 +154,8 @@ iris = {
       trackingTemplateBtn: '#tracking-notification[data-view="false"] .tracking-inner h4',
       variablesTemplateSource: $('#variables-template').html(),
       appSelect: '.template-application',
+      countInput: '.notification-count',
+      waitInput: '.notification-wait',
       template: null,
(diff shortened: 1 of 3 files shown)


src/iris/ui/static/js/iris.js
```diff
@@ -244,8 +246,35 @@ iris = {
...
+
+      $element.parent().siblings('.plan-notification').each(function(_, n) {
+        var $n = $(n);
+        if ($n.attr('data-mode') === 'view') {
+          waits.push($n.children('.notification-wait')[0].title * $n.children('.notification-count')[0].title);
+        } else {
+          waits.push($n.find('.notification-wait')[0].value * $n.find('.notification-count')[0].value);
+        }
+      })
+      stepTime = Math.min(...waits);
```
jrgp: I don't think this spread syntax is supported everywhere: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax

To keep consistent with the rest of the iris JS not using new es6+ features, can we make this use a non-spread equivalent?

PR #441, opened 2018-09-07
Optional notification- Add the ability to mark a plan notification as optional

diegocepedaw:
If notification is optional iris will not email the owner if target is not resolved correctly.


db/schema_0.sql
```diff
@@ -210,6 +210,7 @@ CREATE TABLE `plan_notification` (
   `step` int(11) NOT NULL,
   `template` varchar(255) DEFAULT NULL,
   `target_id` bigint(20),
+  `optional` TINYINT(1) NOT NULL DEFAULT '0',
```
dwang159: I'd prefer this to be BOOLEAN, so we can do True/False in our REST API calls instead of 0/1.

src/iris/api.py
```diff
@@ -278,6 +278,7 @@ def ts_to_sql_datetime(ts):
     `plan_notification`.`step` as `step`,
     `plan_notification`.`repeat` as `repeat`,
     `plan_notification`.`wait` as `wait`,
+    IF(`plan_notification`.`optional` = 1, 'True', 'False') as `optional`,
```
dwang159: BOOLEAN also gets rid of this awkward bit

src/iris/api.py
```diff
@@ -1256,6 +1261,13 @@ def on_post(self, req, resp):
         }
 
         dynamic_indices = set()
+
+        # for backwards copatibility check if optional is not defined and set it to 0 if it isn't
```
dwang159: Instead of iterating through twice, modify the step right before executing the query.

dwang159: Line 1286 is a good place (`step['step'] = index`)

test/e2etest.py
```diff
@@ -876,15 +876,17 @@ def test_post_plan(sample_user, sample_team, sample_template_name):
                     "priority": "low",
                     "wait": 600,
                     "repeat": 0,
-                    "template": sample_template_name
+                    "template": sample_template_name,
+                    "optional": 0
```
dwang159: Make sure there's tests with optional undefined to ensure backwards-compatibility

db/schema_0.sql
diegocepedaw: I defined it as a tinyint for consistency, BOOLEAN should be completely interchangeable https://dev.mysql.com/doc/refman/8.0/en/other-vendor-data-types.html

test/e2etest.py
diegocepedaw: Updated the tests to have some calls with no inclusion of the optional field.

src/iris/api.py
diegocepedaw: Not sure if I can get rid of this since BOOLEAN is also stored as TINYINT(1).

dwang159: This selects the strings "True" and "False". Instead, we should just go with the raw value of 0 or 1 and have the caller cast it.

diegocepedaw: Removed this from select statement and moved the translation from 0 -> "False" and 1 -> "True" to the Jinja template.

src/iris/api.py
diegocepedaw: Changed.

src/iris/api.py
```diff
@@ -1272,6 +1278,10 @@ def on_post(self, req, resp):
                     dynamic = step.get('dynamic_index') is not None
                     step['plan_id'] = plan_id
                     step['step'] = index
+                    # for backwards copatibility check if optional is not defined and set it to 0 if it isn't
+                    if 'optional' not in step:
```
jrgp: You can combine both map operations (the check and the assignment) to one using

    step.setdefault('optional', 0)

src/iris/ui/templates/plan.html
```diff
@@ -324,6 +325,7 @@ <h4><i class="icon-plus icon">+</i> Add application tracking template</h4>
     <p class="item-content" title="{{ secondsToMinutes wait }}"><label class="item-label light" data-toggle="tooltip" title="Minimum time that must elapse before repeating or moving onto the next escalation step.">Wait</label> {{ secondsToMinutes wait }} minutes</p>
     <p class="item-content" title="{{ count }}"><label class="item-label light" data-toggle="tooltip" title="Number of times to execute this notification before escalating to next step">Count</label> {{ count }} times</p>
     <p class="item-content" title="{{ template }}"><label class="item-label light" data-toggle="tooltip" title="Content to use for the notification">Template</label> <a href="/templates/{{template}}">{{ template }}</a></p>
+    <p class="item-content" title="{{ optional }}"> <label class="item-label light" data-toggle="tooltip" title="If set to true, this notification won't generate an alert if it fails to find a valid target.">Optional</label> {{#isEqual optional 1}}True{{else}}False{{/isEqual}}</p>
```
jrgp: You can probably do normal `{{#if optional #}}` instead of explicitly comparing it to 1 as 1 is a truthy value.

src/iris/ui/templates/plan.html
```diff
@@ -371,6 +373,13 @@ <h4><i class="icon-plus icon">+</i> Add application tracking template</h4>
         {{/each}}
       </select>
     </p>
+    <p>
+      <label class="item-label light" data-toggle="tooltip" title="If set to true, this notification won't generate an alert if it fails to find a valid target.">Optional</label>
+      <select class="content-template form-control border-bottom" data-type="optional">
+          <option value="0">False</option>
```
jrgp: When viewing a plan, will this dropdown have the proper option selected?

src/iris/api.py
diegocepedaw: Changed.

src/iris/ui/templates/plan.html
diegocepedaw: Changed.

src/iris/ui/templates/plan.html
diegocepedaw: Yes, the default option selected is False.

jrgp: Why not Yes/No or Enabled/Disabled or On/Off instead of True/False? Exposing boolean names feels like an implementation detail that the user might not care about.

diegocepedaw: There wasn't a specific reason for it to be worded as True/False but you bring up a good point, I think Yes/No might in fact be a bit clearer for the user in this case.

PR #443, opened 2018-09-12
Persist optional value on clone plan

diegocepedaw:


src/iris/ui/templates/plan.html
```diff
@@ -380,8 +380,13 @@ <h4><i class="icon-plus icon">+</i> Add application tracking template</h4>
     <p>
       <label class="item-label light" data-toggle="tooltip" title="If set to true, this notification won't generate an alert if it fails to find a valid target.">Optional</label>
       <select class="content-template form-control border-bottom" data-type="optional">
+        {{# if optional}}
```
jrgp: Typically with dropdowns we want to attach a `selected` attribute to the particular `<option>` tag which is currently selected to make the browser have it be currently selected, rather than letting the currently active item be at the top of the list.

You can see examples of this across other parts of the Iris and oncall codebases. Something like this would work here:

```
<select class="content-template form-control border-bottom" data-type="optional">
    <option value="1" {{# if optional}}selected{{/if}}>Yes</option>
    <option value="0" {{# unless optional}}selected{{/unless}}>No</option>
</select>
```

.gitignore
```diff
@@ -14,3 +14,4 @@ venv
 /src/iris/ui/static/bundles
 .tox
 *.orig
+logs/sender_rpc.access.log
```
dwang159: What is this file?

src/iris/ui/templates/plan.html
dwang159: :+1:

.gitignore
jrgp: iris-sender will likely write that rpc log. We should probably put changes to the .gitignore file in its own PR.

src/iris/ui/templates/plan.html
diegocepedaw: changed

.gitignore
diegocepedaw: removed

PR #445, opened 2018-09-12
Sleep before email retry

dwang159:


src/iris/vendors/iris_smtp.py
```diff
@@ -164,6 +165,8 @@ def send_email(self, message, customizations=None):
                     return None
 
             try:
+                # Sleep to back-off on connection
+                sleep(1)
```
jrgp: I’m wondering if 1 second is too long and might cause problems. Can we make the amount of time we sleep configurable in the vendor config, and not sleep at all if it’s set to 0s?

PR #446, opened 2018-09-12
disallow plan steps that only have optional notifications

diegocepedaw:


src/iris/api.py
```diff
@@ -1331,6 +1337,9 @@ def on_post(self, req, resp):
                         raise HTTPBadRequest('Invalid plan',
                                              'Target not found for step %s' % index)
 
+                if only_optional_flag:
+                    raise HTTPBadRequest('Invalid plan', 'You must have at least one non-optional notification per step. Step %s has none.' % index)
```
jrgp: Looks good. Can we have a unit test added that confirms this behavior?

PR #453, opened 2018-10-12
Literal target notifications

diegocepedaw:
A notification sent using the new "literal_target" role will skip the target unrolling process and instead use the provided target as its destination. This introduces the ability to send notifications to any arbitrary email address (including mailing lists), slack recipient (including channels), and numbers.

### DIFF
--- a/src/iris/bin/sender.py
+++ b/src/iris/bin/sender.py
@@ -782,6 +782,10 @@ def set_target_contact(message):
     # If we already have a destination set (eg incident tracking emails) no-op this
     if 'destination' in message:
         return True
+    if message['role'] == 'literal_target':
+        message['destination'] = message['target']
+        return True
+
     # returns True if contact has been set (even if it has been changed to the fallback). Otherwise, returns False
     try:
         if 'mode' in message or 'mode_id' in message:
@@ -801,7 +805,7 @@ def set_target_contact(message):
             message['destination'] = cursor.fetchone()[0]
             cursor.close()
             connection.close()
-            result = True
+            result = True   
         else:
             # message triggered by incident will only have priority
             result = set_target_contact_by_priority(message)
@@ -1052,13 +1056,14 @@ def fetch_and_send_message(send_queue, vendor_manager):
         message = send_queue.get(True, 4)
     except queue.Empty:
         return
+    blame_str = 'app:%s  target:%s' % (message.get('application', '?'), message.get('target', '?')) 
 
     metrics.incr('send_queue_gets_cnt')
 
     retry_count = message.get('retry_count')
     is_retry = retry_count is not None
     if is_retry and retry_count >= MAX_MESSAGE_RETRIES:
-        logger.warning('Maximum retry count for %s breached')
+        logger.warning('Maximum retry count for %s breached', blame_str)
         return
 
     if not is_retry:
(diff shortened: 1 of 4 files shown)


src/iris/sender/rpc.py
```diff
@@ -89,16 +89,30 @@ def handle_api_notification_request(socket, address, req):
...
-    except IrisRoleLookupException:
-        expanded_targets = None
-    if not expanded_targets:
-        reject_api_request(socket, address, 'INVALID role:target')
-        logger.warn('Dropping OOB message with invalid role:target "%s:%s" from app %s',
-                    role, target, notification['application'])
-        return
+    expanded_targets = None
+    # if role is literal_target skip unrolling
+    if role == 'literal_target':
```
dwang159: Do these checks in `Notifications.on_post()` to avoid needless rpc on bad requests.

src/iris/sender/rpc.py
```diff
@@ -129,10 +143,13 @@ def handle_api_notification_request(socket, address, req):
                        address, role, target, notification['application'],
                        notification.get('priority', notification.get('mode', '?')))
 
-    for _target in expanded_targets:
-        temp_notification = notification.copy()
-        temp_notification['target'] = _target
-        send_funcs['message_send_enqueue'](temp_notification)
+    if role == 'literal_target':
```
dwang159: The `literal_target` pseudo-role is fine for interfacing with users, but once we drop into the API layer we should disambiguate from real roles. I'd prefer an additional member in the notification dict of `unexpanded: True` or something. This way `role` in the rpc and sender code always means a real role.

src/iris/bin/sender.py
```diff
@@ -782,6 +782,10 @@ def set_target_contact(message):
     # If we already have a destination set (eg incident tracking emails) no-op this
     if 'destination' in message:
         return True
+    if message['role'] == 'literal_target':
```
dwang159: Set `destination` outside of the sender, probably in rpc. No need to wait so long, and keeps the expand/unexpand logic out of the sender.

src/iris/bin/sender.py
```diff
@@ -1145,7 +1150,7 @@ def fetch_and_send_message(send_queue, vendor_manager):
     except Exception:
         logger.exception('Failed to send message: %s', message)
         add_mode_stat(message['mode'], None)
-        if message['mode'] == 'email':
+        if message['mode'] == 'email' or message['role'] == 'literal_target':
```
dwang159: Switch to `unexpanded` flag as noted.

src/iris/bin/sender.py
```diff
@@ -1052,13 +1056,14 @@ def fetch_and_send_message(send_queue, vendor_manager):
...
         return
+    blame_str = 'app:%s  target:%s' % (message.get('application', '?'), message.get('target', '?'))
 
     metrics.incr('send_queue_gets_cnt')
 
     retry_count = message.get('retry_count')
     is_retry = retry_count is not None
     if is_retry and retry_count >= MAX_MESSAGE_RETRIES:
-        logger.warning('Maximum retry count for %s breached')
+        logger.warning('Maximum retry count for %s breached', blame_str)
```
jrgp: Why use a separate variable instead of something like:

    logger.warning('Maximum retry count for app: %s target:%s breached', message.get('application', '?'), message.get('target', '?'))

src/iris/bin/sender.py
dwang159: This may also increment task failure if users give bad destinations, which we need to differentiate from actual task failures.

src/iris/sender/rpc.py
```diff
@@ -129,10 +131,14 @@ def handle_api_notification_request(socket, address, req):
                        address, role, target, notification['application'],
                        notification.get('priority', notification.get('mode', '?')))
 
-    for _target in expanded_targets:
-        temp_notification = notification.copy()
-        temp_notification['target'] = _target
-        send_funcs['message_send_enqueue'](temp_notification)
+    if 'unexpanded' in notification:
```
dwang159: Switch from `if 'unexpanded' in` to `if notification.get('unexpanded')`

PR #456, opened 2018-11-06
Add target field to plan search endpoint

diegocepedaw:
This addition allows users to search for all the plans that contain some specific target within their steps.

### DIFF
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -1110,7 +1110,7 @@ class Plans(object):
             except Exception:
                 raise HTTPBadRequest('Invalid provided target.')
 
-            plan_target_join = ' JOIN `plan_notification` ON `plan_notification`.`id` = `plan`.`id` AND `plan_notification`.`target_id` = %s' % plan_target_id
+            plan_target_join = ' JOIN `plan_notification` ON `plan_notification`.`plan_id` = `plan`.`id` AND `plan_notification`.`target_id` = %s' % plan_target_id
             query = query + plan_target_join
 
         where = []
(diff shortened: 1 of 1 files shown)


src/iris/api.py
```diff
@@ -1092,6 +1098,21 @@ def on_get(self, req, resp):
 
         query = plan_query % ', '.join(plan_columns[f] for f in fields)
 
+        connection = db.engine.raw_connection()
+        cursor = connection.cursor(db.ss_dict_cursor)
+
+        plan_target = req.params.pop('target', None)
```
dwang159: You should be able to integrate this into `gen_where_filter_clause`. That gets us `startswith`, `contains`, and `endswith` for free. We'll probably want to use `contains` for integrating this into the UI anyway.

src/iris/api.py
```diff
@@ -1092,6 +1098,21 @@ def on_get(self, req, resp):
...
+        # search for plans which have steps that target a specific user
+        if plan_target is not None:
+            try:
+                cursor.execute('SELECT `id` FROM `target` WHERE `name` = %s', plan_target)
+                plan_target_id = cursor.fetchone()['id']
+            except Exception:
+                raise HTTPBadRequest('Invalid provided target.')
```
dwang159: Instead of this, raise `HTTPBadRequest` if `cursor.fetchone()` returns `None`. We shouldn't hide possible DB errors with a 400. Although this may go away if you use `gen_where_filter_clause`

src/iris/api.py
diegocepedaw: implemented

src/iris/api.py
diegocepedaw: changed to use same dbquery as all the search calls

PR #471, opened 2019-01-08
Add ability to comment on incidents

dwang159:
Also, add comments to retention script to handle archiving incidents with comments.

### DIFF
--- a/src/iris/ui/static/js/iris.js
+++ b/src/iris/ui/static/js/iris.js
@@ -1466,24 +1466,29 @@ iris = {
     },
     addComment: function() {
       var self = this,
+          commentContainer = $(self.data.addCommentContainer),
           comment = {
             author: window.appData.user,
             content: $('#comment-body').val()
           };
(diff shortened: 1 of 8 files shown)


src/iris/ui/static/js/iris.js
```diff
@@ -1438,11 +1444,46 @@ iris = {
...
+    hideComment: function() {
+      $('#comment-body').val('');
+      $(this.data.addCommentContainer).hide();
+      $(this.data.showAddCommentBtn).show();
+    },
+    addComment: function() {
+      var self = this,
+          comment = {
+            author: window.appData.user,
+            content: $('#comment-body').val()
```
jrgp: Can we verify the given body isn't an empty string/a string with just whitespace before we allow the form to be submitted?

jrgp:
Changes in the latest commit look good. Looks like we have a problem with the code coverage causing the CI to fail though.

PR #483, opened 2019-01-29
Resolved

diegocepedaw:
Adds the 'resolved' attribute to incidents. The idea is that resolved will allow for visibility into wether a claimed incident is still being worked on or if the problem has been fixed and the incident is resolved.


test/e2etest.py
```diff
@@ -1298,6 +1298,15 @@ def test_post_incident(sample_user, sample_team, sample_application_name, sample
     assert re.status_code == 400
     assert re.json()['title'] == 'Invalid claim: no matching owner'
 
+    # Test reslovling
```
jrgp: Resolving is typo'd

src/iris/ui/static/js/iris.js
```diff
@@ -1548,6 +1552,30 @@ iris = {
...
+        method: 'POST',
+        contentType: 'application/json'
+      }).done(function(){
+        self.getIncident(incidentId).done(function(){
+          var message = owner ? 'Incident ' + incidentId + ' marked as resolved.' : 'Incident ' + incidentId + ' marked as unresolved.';
+          iris.createAlert(message, 'success');
+        });
+      }).fail(function(){
+        iris.createAlert('Failed to modify incident', 'danger');
+      }).always(function(){
```
jrgp: Do we need this always callback? Looks like it only contains commented code.

diegocepedaw: We do not, I just forgot to remove it.

src/iris/utils.py
```diff
@@ -312,6 +312,66 @@ def claim_incidents_from_batch_id(batch_id, owner):
...
+      SELECT `target`.`name`
+      FROM `incident`
+      LEFT JOIN `target` ON `target`.`id` = `incident`.`owner_id`
+      WHERE `incident`.`id` = %s
+    ''', (incident_id, ))
+    result = cursor.fetchone()
+    previous_owner = result[0] if result else None
+    cursor.close()
+
+    resolved = 1 if owner else 0
```
dwang159: Should this be called `unclaimed`?

src/iris/utils.py
```diff
@@ -312,6 +312,66 @@ def claim_incidents_from_batch_id(batch_id, owner):
...
+
+    max_retries = 3
+
+    for i in xrange(max_retries):
+        cursor = connection.cursor()
+        try:
+            cursor.execute('''UPDATE `incident`
+                               SET `incident`.`updated` = %(updated)s,
+                                   `incident`.`resolved` = %(resolved)s,
+                                   `incident`.`owner_id` = (SELECT `target`.`id` FROM `target` WHERE `target`.`name` = %(owner)s AND `type_id` = (SELECT `id` FROM `target_type` WHERE `name` = 'user'))
```
dwang159: I don't think clicking "resolved" should change the incident owner. Owner should match claimer, and we can add a new field if we want to track resolver. We can get rid of the owner_id update this way.

src/iris/utils.py
```diff
@@ -312,6 +312,66 @@ def claim_incidents_from_batch_id(batch_id, owner):
...
+            logger.exception('Failed running resolve query. (Try %s/%s)', i + 1, max_retries)
+        finally:
+            cursor.close()
+
+        sleep(.2)
+
+    cursor = connection.cursor()
+
+    try:
+        cursor.execute('UPDATE `message` SET `resolved` = 1 WHERE `incident_id`= %s', (incident_id, ))
```
dwang159: Delete after removing `message.resolved` column

PR #484, opened 2019-01-31
grafana webhook support

wleese:
Adds a new webhook to natively support using Iris as a notification channel in Grafana, using the 'webhook' type.

### DIFF
--- a/src/iris/webhooks/grafana.py
+++ b/src/iris/webhooks/grafana.py
@@ -1,0 +1,87 @@
+from __future__ import absolute_import
+
+import datetime
+import logging
+import ujson
+from falcon import HTTP_201, HTTPBadRequest, HTTPInvalidParam
+
+from iris import db
+
+logger = logging.getLogger(__name__)
+
+
+class grafana(object):
+    allow_read_no_auth = False
+
+    def validate_post(self, body):
+        if not all(k in body for k in("ruleName", "state", "message")):
+            raise HTTPBadRequest('missing ruleName, state and/or message attributes')
+
+    def create_context(self, body):
+        context_json_str = ujson.dumps(body)
+        if len(context_json_str) > 65535:
+            logger.warn('POST to grafana exceeded acceptable size')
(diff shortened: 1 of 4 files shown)


test/test_webhook_grafana.py
```diff
@@ -0,0 +1,57 @@
...
+        }],
+        "imageUrl": "http://grafana.org/assets/img/blog/mixed_styles.png",
+        "message": "Someone is testing the alert notification within grafana.",
+        "ruleId": 0,
+        "ruleName": "Test notification",
+        "ruleUrl": "https://grafana.org/",
+        "state": "alerting",
+        "title": "[Alerting] Test notification"
+    }
+    try:
```
jrgp: Why not use something like the following to ensure the exception does get raised? With the current method of catching It we don't know for sure if it was raised or not.

```
with pytest.raises(HTTPBadRequest):
    grafana_webhook.validate_post(fake_post)
```

src/iris/webhooks/grafana.py
```diff
@@ -0,0 +1,87 @@
...
+            incident_id = session.execute(
+                '''INSERT INTO `incident` (`plan_id`, `created`, `context`,
+                                           `current_step`, `active`, `application_id`)
+                   VALUES (:plan_id, :created, :context, 0, :active, :application_id)''',
+                data).lastrowid
+
+            session.commit()
+            session.close()
+
+        resp.status = HTTP_201
```
jrgp: If we're adding the Location header, this response becomes a redirect, so we should do a 301 or 302 rather than a 2xx

wleese:
I'll open up a new PR with the suggested fixes, thanks!

test/test_webhook_grafana.py
wleese: https://github.com/linkedin/iris/pull/486

src/iris/webhooks/grafana.py
wleese: https://github.com/linkedin/iris/pull/486

PR #485, opened 2019-02-02
Add QR code front end route for iris mobile integration

diegocepedaw:


src/iris/ui/__init__.py
```diff
@@ -399,6 +421,9 @@ def init(config, app):
     auth_module = config.get('auth', {'module': 'iris.ui.auth.noauth'})['module']
     auth = importlib.import_module(auth_module)
     auth_manager = getattr(auth, 'Authenticator')(config)
+    qr_base_url = config['qr_base_url']
```
dwang159: We should change this (and the qr endpoint) to be optional, since it has no meaning if mobile isn't configured.

diegocepedaw: Implemented

diegocepedaw: Makes sense, implemented.

PR #490, opened 2019-02-13
Title variable functionality 

diegocepedaw:
Allow users to define an application variable as a 'title variable' this variable if defined  will be used as the incident title on the mobile app.

### DIFF
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -2508,8 +2508,6 @@ class Application(object):
 
             if 'mobile_template' not in data:
                 raise HTTPBadRequest('mobile_template must be specified')
-            if 'title_variable' not in data:
-                raise HTTPBadRequest('title_variable must be specified')
 
             new_variables = data.get('variables')
             if not isinstance(new_variables, list):
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -162,6 +162,7 @@ incident_columns = {
     'created': 'UNIX_TIMESTAMP(`incident`.`created`) as `created`',
     'owner': '`target`.`name` as `owner`',
     'current_step': '`incident`.`current_step` as `current_step`',
+    'title_variable_name': '`template_variable`.`name` as `title_variable_name`'
 }
 
 incident_filters = {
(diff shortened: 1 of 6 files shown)


src/iris/api.py
```diff
@@ -2514,12 +2520,30 @@ def on_put(self, req, resp, app_name):
...
+            elif title_variable and title_variable not in new_variables:
+                raise HTTPBadRequest('title variable is invalid')
+
             kill_variables = existing_variables - new_variables
 
-            for variable in new_variables - existing_variables:
-                session.execute('''INSERT INTO `template_variable` (`application_id`, `name`)
-                                VALUES (:application_id, :variable)''',
-                                {'application_id': app['id'], 'variable': variable})
+            # insert new variables and update the value of title_variable for existing variables
```
dwang159: We can reduce DB queries by keeping the old variable insertion query, then doing a `UPDATE template_variable SET title_variable=(CASE WHEN name=:title_val THEN TRUE ELSE FALSE) WHERE application_id = (SELECT id from application where name=:app)` (or something similar, I didn't test this).

src/iris/api.py
```diff
@@ -162,6 +162,7 @@ def ts_to_sql_datetime(ts):
     'created': 'UNIX_TIMESTAMP(`incident`.`created`) as `created`',
     'owner': '`target`.`name` as `owner`',
     'current_step': '`incident`.`current_step` as `current_step`',
+    'title_variable_name': '`template_variable`.`name` as `title_variable_name`'
```
dwang159: Looks like you're adding title_variable_name into the fields that we get from a query to `/incidents`. I like this idea, but let's take it a step further and actually inject a new field into the JSON that we return, like `title`. Then, the API can be responsible for grabbing the title from the context (if it exists), and the clients can just use that field without needing to dig into the context.

diegocepedaw: /incidents actually already returns the title information in the "title_variable_name" variable. It's defined in the incident_columns.

src/iris/api.py
diegocepedaw: That makes sense, I'll make a change to that effect.

src/iris/api.py
dwang159: Sure, but I mean that we can have the API abstract away the title variable name and just return the incident title. E.g returning `{id: 123, title: "5xx Errors above threshold"...}` instead of `{id: 123, title_variable_name: "name", context: {"name": "5xx Errors above threshold"}}`.

diegocepedaw: Changed.

src/iris/api.py
```diff
@@ -574,9 +577,15 @@ def ts_to_sql_datetime(ts):
...
 
 
-def stream_incidents_with_context(cursor):
+def stream_incidents_with_context(cursor, title=False):
     for row in cursor:
         row['context'] = ujson.loads(row['context'])
+        if title:
+            title_variable_name = row['title_variable_name']
+            if title_variable_name:
+                row['title'] = row['context'][title_variable_name]
```
dwang159: Let's do a `.get(title_variable_name)` just in case

src/iris/api.py
```diff
@@ -1429,7 +1438,10 @@ def on_get(self, req, resp):
         cursor.execute(query, sql_values)
 
         if 'context' in fields:
-            payload = ujson.dumps(stream_incidents_with_context(cursor))
+            if 'title_variable_name' in fields:
```
dwang159: Suggest renaming the field name to just `title`

src/iris/api.py

PR #492, opened 2019-02-19
add total_high_priority stat to global stats

diegocepedaw:


src/iris/app_stats.py
```diff
@@ -172,7 +172,8 @@ def calculate_global_stats(connection, cursor, fields_filter=None):
                                                                 ORDER BY time_to_claim) as time_to_claim
                                                         WHERE (SELECT @row_id := @row_id + 1)
                                                         BETWEEN @incident_count/2.0 AND @incident_count/2.0 + 1)''',
-        'total_applications': 'SELECT COUNT(*) FROM `application` WHERE `auth_only` = FALSE'
+        'total_applications': 'SELECT COUNT(*) FROM `application` WHERE `auth_only` = FALSE',
+        'total_high_priority': 'SELECT COUNT(DISTINCT incident.id) FROM incident JOIN message ON message.incident_id = incident.id WHERE incident.created > NOW() - INTERVAL 2 WEEK AND (priority_id = 8 OR priority_id = 17)'
```
dwang159: So I know I gave you that query, but on second thought we should change the last bit to `priority_id IN (SELECT id FROM priority WHERE name IN ('high', 'urgent'))`.

Also, consider naming `high_priority_incidents_last_2_weeks`, since the stat isn't a total.

diegocepedaw: Sounds good, I'll make the change

PR #497, opened 2019-03-13
Bump twilio python to modern version

dwang159:
Update the way we handle proxy settings
Timeouts now require superclassing and replacing Twilio's http client.

### DIFF
--- a/src/iris/vendors/iris_twilio.py
+++ b/src/iris/vendors/iris_twilio.py
@@ -4,8 +4,8 @@
 from iris.constants import SMS_SUPPORT, CALL_SUPPORT
 from iris.plugins import find_plugin
 from iris.custom_import import import_custom_module
-from twilio.rest import TwilioRestClient
-from twilio.rest.resources import Connection
+from twilio.rest import Client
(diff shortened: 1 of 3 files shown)


src/iris/vendors/iris_twilio.py
```diff
@@ -20,15 +20,15 @@ class iris_twilio(object):
...
-                                      self.config['proxy']['port'],
-                                      proxy_rdns=True)
+            self.proxy_client = TwilioHttpClient()
+            proxy = '%s:%s' % (self.config['proxy']['host'], self.config['proxy']['port'])
+            self.proxy_client.session.proxies = {'https': proxy, 'http': proxy}
         self.modes = {
             SMS_SUPPORT: self.send_sms,
             CALL_SUPPORT: self.send_call,
         }
-        self.timeout = config.get('timeout', 10)
```
jrgp: Is there a way of specifying the timeout with this new client?

dwang159: Thought of a better way to preserve timeouts. Updated.

jrgp: LGTM

PR #503, opened 2019-03-28
Py3 internal client

diegocepedaw:


src/iris/client.py
```diff
@@ -12,16 +12,20 @@
 class IrisAuth(requests.auth.AuthBase):
     def __init__(self, app, key):
         self.header = 'hmac %s:' % app
-        self.HMAC = hmac.new(key, '', hashlib.sha512)
+        key = key if isinstance(key, bytes) else key.encode('utf8')
+        self.HMAC = hmac.new(key, b'', hashlib.sha512)
 
     def __call__(self, request):
         HMAC = self.HMAC.copy()
```
dwang159: This probably needs tweaking, since `'%s' % b'asdf' == "b'asdf'"`. See https://github.com/houqp/iris-python-client/blob/9ff45ce64fddac342b38ec771d7f9181edf99c59/src/irisclient/__init__.py#L30

diegocepedaw: Copied, https://github.com/houqp/iris-python-client/blob/master/src/irisclient/__init__.py

PR #504, opened 2019-04-02
Encode results of ldap query to utf-8

diegocepedaw:


src/iris/bin/sync_targets.py
```diff
@@ -309,7 +309,16 @@ def get_ldap_lists(l, search_strings, parent_list=None):
                              filterstr=filterstr)
         rtype, rdata, rmsgid, serverctrls = l.result3(msgid, timeout=ldap_timeout, resp_ctrl_classes=known_ldap_resp_ctrls)
 
-        results |= {(data[search_strings['list_cn_field']][0], data[search_strings['list_name_field']][0]) for (dn, data) in rdata}
+        for (dn, data) in rdata:
+            cn_field = data[search_strings['list_cn_field']][0]
+            name_field = data[search_strings['list_name_field']][0]
+
+            if isinstance(cn_field, bytes):
```
jrgp: Are these fields sometimes str and sometimes bytes? Or are they always bytes? If so, why not skip all the isinstance checks and just run `.decode('utf-8')` unconditionally?

diegocepedaw: Some fields in the LDAP are represented as text (distinguished names, relative distinguished names, attribute names, queries) and some are represented as bytes. I could unconditionally run .decode('utf-8') on attribute values but since it is a call to a downstream we don't own I opted to check all of them for consistency and to avoid issues if it ever changes.

PR #507, opened 2019-04-04
Sync oncall-api team id and automatically propagate team renaming

diegocepedaw:
I'd like to write some tests before it's ready but I'd appreciate some feedback on this in the meantime.

### DIFF
--- a/db/schema_0.sql
+++ b/db/schema_0.sql
@@ -320,7 +320,7 @@ CREATE TABLE `oncall_team` (
   `name` varchar(255) NOT NULL,
   `oncall_team_id` bigint(20) NOT NULL,
   PRIMARY KEY (`target_id`),
-  CONSTRAINT `oncall_team_ibfk_1` FOREIGN KEY (`target_id`) REFERENCES `target` (`id`) ON DELETE CASCADE ON DELETE CASCADE ON UPDATE CASCADE,
+  CONSTRAINT `oncall_team_ibfk_1` FOREIGN KEY (`target_id`) REFERENCES `target` (`id`) ON DELETE CASCADE ON UPDATE CASCADE
 ) ENGINE=InnoDB DEFAULT CHARSET=latin1;
 /*!40101 SET character_set_client = @saved_cs_client */;
 
--- a/db/schema_0.sql
+++ b/db/schema_0.sql
@@ -307,6 +307,23 @@ CREATE TABLE `target` (
 ) ENGINE=InnoDB DEFAULT CHARSET=latin1;
 /*!40101 SET character_set_client = @saved_cs_client */;
 
+
+--
+-- Table structure for table `oncall_team`
+--
+
+DROP TABLE IF EXISTS `oncall_team`;
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!40101 SET character_set_client = utf8 */;
+CREATE TABLE `oncall_team` (
+  `target_id` bigint(20) NOT NULL,
+  `name` varchar(255) NOT NULL,
+  `oncall_team_id` bigint(20) NOT NULL,
+  PRIMARY KEY (`target_id`),
+  CONSTRAINT `oncall_team_ibfk_1` FOREIGN KEY (`target_id`) REFERENCES `target` (`id`) ON DELETE CASCADE ON DELETE CASCADE ON UPDATE CASCADE,
+) ENGINE=InnoDB DEFAULT CHARSET=latin1;
+/*!40101 SET character_set_client = @saved_cs_client */;
+
 --
 -- Table structure for table `target_application_mode`
 --
--- a/db/schema_0.sql
+++ b/db/schema_0.sql
@@ -319,6 +319,7 @@ CREATE TABLE `oncall_team` (
   `target_id` bigint(20) NOT NULL,
   `oncall_team_id` bigint(20) NOT NULL,
   PRIMARY KEY (`target_id`),
+  UNIQUE KEY `oncall_team_id_idx` (`oncall_team_id`),
   CONSTRAINT `oncall_team_ibfk_1` FOREIGN KEY (`target_id`) REFERENCES `target` (`id`) ON DELETE CASCADE ON UPDATE CASCADE
 ) ENGINE=InnoDB DEFAULT CHARSET=latin1;
 /*!40101 SET character_set_client = @saved_cs_client */;
--- a/db/schema_0.sql
+++ b/db/schema_0.sql
@@ -317,7 +317,6 @@ DROP TABLE IF EXISTS `oncall_team`;
 /*!40101 SET character_set_client = utf8 */;
 CREATE TABLE `oncall_team` (
   `target_id` bigint(20) NOT NULL,
-  `name` varchar(255) NOT NULL,
   `oncall_team_id` bigint(20) NOT NULL,
   PRIMARY KEY (`target_id`),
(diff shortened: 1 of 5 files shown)


db/schema_0.sql
```diff
@@ -307,6 +307,23 @@ CREATE TABLE `target` (
...
+--
+-- Table structure for table `oncall_team`
+--
+
+DROP TABLE IF EXISTS `oncall_team`;
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!40101 SET character_set_client = utf8 */;
+CREATE TABLE `oncall_team` (
+  `target_id` bigint(20) NOT NULL,
+  `name` varchar(255) NOT NULL,
```
jrgp: should name be unique? that field should at least have some index as it's being used in WHERE clauses

src/iris/bin/sync_targets.py
```diff
@@ -163,12 +172,18 @@ def sync_from_oncall(config, engine, purge_old_users=True):
         logger.warning('No users found. Bailing.')
         return
 
-    oncall_team_names = fetch_teams_from_oncall(oncall)
+    # get teams from oncall-api and separate the list of touples into two lists of name and ids
```
jrgp: touples -> tuples typo

src/iris/bin/sync_targets.py
```diff
@@ -163,12 +172,18 @@ def sync_from_oncall(config, engine, purge_old_users=True):
...
+    oncall_team_response = list(zip(*fetch_teams_from_oncall(oncall)))
+    oncall_team_names = oncall_team_response[0]
```
jrgp: If fetch_teams_from_oncall returns no result, the [0] and [1] here will raise IndexErrors

src/iris/bin/sync_targets.py
```diff
@@ -263,25 +278,88 @@ def sync_from_oncall(config, engine, purge_old_users=True):
...
+    possibly_mismatched_names = iris_team_names - oncall_team_names
+
+    # find teams in the iris database whose names have changed
+    for name in possibly_mismatched_names:
+        if iris_db_oncall_team_name_id_dict.get(name) in oncall_team_ids:
+            # rename team to new name
+            target_id_to_rename = iris_target_name_id_dict[name]
+            # get the oncall response name using the oncall_team team_id
+            new_name = oncall_response_dict_id_key[iris_db_oncall_team_name_id_dict[name]]
+            logger.info('Renaming team %s to %s' % (name, new_name))
```
jrgp: the python logging functions automatically run % against the arguments you give for you, so this could be simplified to:

    logger.info('Renaming team %s to %s', name, new_name)

This applies to all of the other logger calls in this diff

src/iris/bin/sync_targets.py
```diff
@@ -336,11 +423,15 @@ def get_ldap_list_membership(l, search_strings, list_name):
...
                              serverctrls=[req_ctrl],
                              attrlist=[search_strings['user_mail_field']],
-                             filterstr=search_strings['user_membership_filter'] % escape_filter_chars(list_name.decode('utf-8'))
+                             filterstr=search_strings['user_membership_filter'] % escape_filter_chars(list_name)
                              )
         rtype, rdata, rmsgid, serverctrls = l.result3(msgid, timeout=ldap_timeout, resp_ctrl_classes=known_ldap_resp_ctrls)
 
-        results |= {data[1].get(search_strings['user_mail_field'], [None])[0] for data in rdata}
+        for data in rdata:
+            member = data[1].get(search_strings['user_mail_field'], [None])[0]
```
jrgp: From this it looks like if the user_mail_field field does not exist, we'll get back a None, and end up inserting a `None` into our results set. Do we want that?

test/test_sync_targets.py
```diff
@@ -0,0 +1,103 @@
...
+    assert iris_team_names - {'demo_team', 'foo_team', 'bar_team'} == set({})
+
+    # return only bar team. This simulates deleting demo and foo team from oncall
+    mocker.patch('iris.bin.sync_targets.fetch_teams_from_oncall').return_value = [["bar_team", 10003]]
+    sync_from_oncall(dummy_configs, engine)
+
+    iris_oncall_team_ids = {oncall_team_id for (oncall_team_id, ) in engine.execute('''SELECT `oncall_team_id` FROM `oncall_team`''')}
+    iris_team_names = {name for (name, ) in engine.execute('''SELECT `name` FROM `target` WHERE `type_id` = 35''')}
+    assert iris_oncall_team_ids == {10003}
+    assert iris_team_names - {'bar_team'} == set({})
```
jrgp: `set({})` can be simplified to `set()`

test/test_sync_targets.py
```diff
@@ -0,0 +1,103 @@
...
+    mocker.patch('iris.bin.sync_targets.fetch_teams_from_oncall').return_value = sample_oncall_teams_response
+    mocker.patch('iris.bin.sync_targets.fetch_users_from_oncall').return_value = sample_oncall_users_response
+    mocker.patch('iris.metrics.stats')
+
+    engine = get_db_engine_from_config(sample_db_config)
+    dummy_configs = {'oncall-api': 'foo'}
+    sync_from_oncall(dummy_configs, engine)
+
+    iris_oncall_team_ids = {oncall_team_id for (oncall_team_id, ) in engine.execute('''SELECT `oncall_team_id` FROM `oncall_team`''')}
+    iris_team_names = {name for (name, ) in engine.execute('''SELECT `name` FROM `target` WHERE `type_id` = 35''')}
```
dwang159: What is type_id 35? Probably should switch to a subquery

src/iris/bin/sync_targets.py
```diff
@@ -102,6 +102,14 @@ def prune_target(engine, target_name, target_type):
     else:
         metrics.incr('others_purged')
 
+    if target_type == 'team':
```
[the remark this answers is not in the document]
diegocepedaw: I don't think we have a reason to keep old oncall_team data even after a soft delete I don't think there's much of a downside to cleaning up the garbage.

src/iris/bin/sync_targets.py
```diff
@@ -244,44 +264,107 @@ def sync_from_oncall(config, engine, purge_old_users=True):
...
-    logger.info('Teams to insert (%d)' % len(teams_to_insert))
-    for t in teams_to_insert:
-        logger.info('Inserting %s' % t)
+# rename all mismatching target names
+
+    iris_db_oncall_team_name_id_dict = {name: team_id for name, team_id in engine.execute('''SELECT target.name, oncall_team.oncall_team_id FROM `target` JOIN `oncall_team` ON oncall_team.target_id = target.id''')}
+    possibly_mismatched_names = iris_team_names - oncall_team_names
+
+    # find teams in the iris database whose names have changed
+    for name in possibly_mismatched_names:
```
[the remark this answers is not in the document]
diegocepedaw: Good point. I changed the logic to rely on oncall_team_id to avoid this type of conflict.

db/schema_0.sql
```diff
@@ -307,6 +307,22 @@ CREATE TABLE `target` (
...
+--
+-- Table structure for table `oncall_team`
+--
+
+DROP TABLE IF EXISTS `oncall_team`;
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!40101 SET character_set_client = utf8 */;
+CREATE TABLE `oncall_team` (
+  `target_id` bigint(20) NOT NULL,
+  `oncall_team_id` bigint(20) NOT NULL,
```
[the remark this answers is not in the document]
diegocepedaw: There shouldn't be any problems with it being unique. Even in the swap case it should be fine since the team_id won't actually change at all.

PR #515, opened 2019-04-26
Stats UI and 7 week lookback

diegocepedaw:
Add a a frontend for all the stats, create high priority incident stats endpoint, return all stats with 7 week lookback


src/iris/api.py
```diff
@@ -4243,18 +4243,68 @@ def on_get(self, req, resp):
...
+
+    def on_get(self, req, resp):
+
+        connection = db.engine.raw_connection()
+        cursor = connection.cursor()
+        stats = {}
+        if self.real_time:
+            stats = app_stats.calculate_high_priority_incidents(connection, cursor)
+        else:
+            cursor.execute('SELECT timestamp, name, value FROM application_stats JOIN application on application_stats.application_id = application.id WHERE statistic = "high_priority_incidents_last_week" ORDER BY value DESC')
```
jrgp: can this query be broken into multiple lines so it becomes easier to read?

src/iris/api.py
```diff
@@ -4243,18 +4243,68 @@ def on_get(self, req, resp):
...
+            if cursor.rowcount == 0:
+                logger.exception('Error retrieving hpi stats from db')
+                raise HTTPInternalServerError('Error retrieving stats from db')
```
jrgp: If this raises, the connection and cursor close lines will never get called, leading to us potentially leaking connetions.

Also, why is a rowcount of 0 worthy of a 5xx? 5xx usually refers to things actually being broken / database exceptions being raised

src/iris/api.py
diegocepedaw: I'll split it up

src/iris/api.py
diegocepedaw: The reason it's raising a 5xx is because unless no apps exist in the database the rowcount should never be 0 here. If the rowcount is 0 it means that the app_stats script failed and the stats tables are in a bad state.

src/iris/app_stats.py
```diff
@@ -84,116 +98,180 @@ def calculate_app_stats(app, connection, cursor, fields_filter=None):
...
+def calculate_high_priority_incidents(connection, cursor):
+
+    high_priority_query = '''SELECT application.name, COUNT(DISTINCT incident.application_id, incident.id) AS incident_count FROM incident JOIN application ON application.id = incident.application_id JOIN message ON message.incident_id = incident.id WHERE incident.created > %(date_variable)s - INTERVAL 1 WEEK AND incident.created < %(date_variable)s AND priority_id IN ( SELECT id FROM priority WHERE name IN ('high', 'urgent') ) GROUP BY incident.application_id ORDER BY incident_count DESC'''
+
+    stats = {}
+
+    for delta in range(0, 7):
+        now = datetime.utcnow() - timedelta(weeks=delta)
+        formatted_date = now.strftime('%Y-%m-%d %H:%M:%S')
+        stats[formatted_date] = []
```
dwang159: When we're passing datetimes around, we should always use Unix timestamps to avoid timezone shenanigans. Converting to a string may/may not be necessary when talking to the DB, you'll need to double check that. For the stats variable, we should keep it in Unix time, then convert to pretty print in the frontend if needed.

src/iris/app_stats.py
```diff
@@ -84,116 +98,180 @@ def calculate_app_stats(app, connection, cursor, fields_filter=None):
...
+            if result:
+                result = result[-1]
+            if result is None:
+                result = -1
+
+            logger.info('Stats query %s took %s seconds', key, round(time.time() - start, 2))
+
+            # {statistic : {timestamp: value, timestamp: value}}
+            if stats.get(key):
+                stats[key][formatted_date] = result
```
dwang159: Same here for the date stuff

src/iris/api.py
```diff
@@ -4265,26 +4315,41 @@ def __init__(self, config):
         self.real_time = cfg.get('real_time', True)
 
     def on_get(self, req, resp, app_name):
-        app = cache.applications.get(app_name)
```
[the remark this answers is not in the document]
diegocepedaw: The stats endpoint is such a low qps endpoint that I didn't see much justification for using the application cache here. Plus the cache is currently case sensitive which is not ideal for checking the application name in the url aginst.

src/iris/api.py
```diff
@@ -4243,18 +4243,68 @@ def on_get(self, req, resp):
...
+                    stats[row[0]] = {}
+                    stats[row[0]][row[2]] = row[1]
 
         cursor.close()
         connection.close()
         resp.status = HTTP_200
         resp.body = ujson.dumps(stats, sort_keys=True)
 
 
+class HighPriorityIncidents(object):
```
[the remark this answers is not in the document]
diegocepedaw: I considered handling as /stats/{statistic} but since high priority incidents per week was the only statistic highlighted as being of particular importance for weekly reporting I decided to avoid naming conflicts and just make it its own endpoint.

src/iris/app_stats.py
```diff
@@ -28,44 +29,57 @@ def calculate_app_stats(app, connection, cursor, fields_filter=None):
...
+        query_data = {'application_id': app['id'], 'date_variable': formatted_date}
+
+        for key in fields:
+            start = time.time()
+            cursor.execute(queries[key], query_data)
+            result = cursor.fetchone()
+            if result:
+                result = result[-1]
+            if result is None:
+                result = -1
```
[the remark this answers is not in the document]
diegocepedaw: The schema originally defined the value field as NOT NULL but considering how I've changed the behavior it probably makes more sense to change that to allow nulls and change the -1s to None.

src/iris/app_stats.py
```diff
@@ -84,116 +68,226 @@ def calculate_app_stats(app, connection, cursor, fields_filter=None):
...
+                            GROUP BY `mode_id`) `counts`
+                        JOIN `mode` ON `mode`.`id` = `counts`.`mode_id`''',
+                       query_data)
+
+        for row in cursor:
+            key = 'total_%s_sent_last_week' % row[0]
+            result = row[1]
+
+            if stats.get(key):
+                stats[key].appennd({unix_date: result})
```
dwang159: `appennd` is a typo here and on line 169. We should fix this typo, make sure the rest of the script has been tested thoroughly, and try to patch the gaps in our e2e testing to make sure we don't miss things like this.

src/iris/api.py
```diff
@@ -4253,12 +4253,12 @@ def on_get(self, req, resp):
             stats = {}
 
             for row in cursor:
-                # {statistic : {timestamp: value, timestamp: value}}
+                # format: {statistic : [{timestamp: value}, {timestamp: value}]}
```
dwang159: Why the change in format? I don't particularly care for one over the other, but curious why it changed.

src/iris/bin/app_stats.py
```diff
@@ -44,22 +45,30 @@ def set_global_stats(stats, connection, cursor):
...
-            cursor.execute('''INSERT INTO `global_stats` (`statistic`, `value`, `timestamp`)
-                            VALUES (%s, %s, %s)
-                            ON DUPLICATE KEY UPDATE `value`= %s, `timestamp` = %s''',
-                           (stat, val, timestamp, val, timestamp))
+    # format: {statistic : [{timestamp: value}, {timestamp: value}]}
+    for stat, entry_list in stats.items():
+        for stat_entry in entry_list:
+            for timestamp, val in stat_entry.items():
+                # format timestamp into datetime to store in mysql
+                timestamp = datetime.utcfromtimestamp(timestamp).strftime('%Y-%m-%d %H:%M:%S')
```
dwang159: For future reference, `FROM_UNIXTIME()` will do this in SQL, so you don't have to mess with datetime and timestring formats.

test/e2etest.py
```diff
@@ -2160,38 +2160,14 @@ def test_healthcheck():
 def test_stats():
     re = requests.get(base_url + 'stats')
     assert re.status_code == 200
-    data = re.json()
```
dwang159: Speaking of testing, are there any validation checks we can do on the data outside of just checking for HTTP 200? Something like what's deleted here, to make sure we're returning data that looks like it's correct. Doesn't need to be as extensive; spot checking is probably fine.

src/iris/api.py
diegocepedaw: I changed it to use lists so it would preserve sorted order when rendered by the browser.

PR #518, opened 2019-05-07
Change calls to Chinese numbers to sms

diegocepedaw:
Due to calling limitations for china change calls made to Chinese numbers to use sms instead. https://support.twilio.com/hc/en-us/articles/360016488474-Calling-Limitations-to-China

### DIFF
--- a/src/iris/bin/sender.py
+++ b/src/iris/bin/sender.py
@@ -1441,6 +1441,18 @@ def sender_shutdown():
     os._exit(0)
 
 
+def check_call_restrictions(message):
+    # check for known corner cases in message delivery and correct accordingly
+
+    # due to calling restrictions to china override mode and send as sms instead
+    if message['destination'].startswith('+86'):
+        message['mode'] = 'sms'
+        # this message will appear before the contents of the sms template
+        message['body'] = '''Due to legal restrictions we are unable to deliver calls to China. Please check Iris for complete incident details and change your settings to use sms, slack, or email instead - '''
+
+    return message['mode'], message['body']
+
+
 def message_send_enqueue(message):
 
     # If this message has an ID, avoid queueing it twice, assuming it's not a retry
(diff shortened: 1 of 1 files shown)


src/iris/bin/sender.py
```diff
@@ -1462,6 +1474,12 @@ def message_send_enqueue(message):
 
     message_mode = message.get('mode')
     if message_mode and message_mode in per_mode_send_queues:
+        # check for known corner case limitations and correct accordingly
+        if message['mode'] == 'call':
```
dwang159: When we make changes like this, we have an `auditlog.message_change()` function that we call to indicate that a change has occurred.

src/iris/bin/sender.py
```diff
@@ -1441,6 +1441,18 @@ def sender_shutdown():
     os._exit(0)
 
 
+def check_call_restrictions(message):
+    # check for known corner cases in message delivery and correct accordingly
+
+    # due to calling restrictions to china override mode and send as sms instead
+    if message['destination'].startswith('+86'):
+        message['mode'] = 'sms'
```
dwang159: We update message twice, once in the helper function and once in `message_send_enqueue`. Not sure there's enough code to bother with a helper function if we move the edits out, but `check_call_restrictions` is a misleading name for a function with side effects. Recommend just putting all this in message_send_enqueue.

src/iris/bin/sender.py
```diff
@@ -1462,6 +1474,12 @@ def message_send_enqueue(message):
 
     message_mode = message.get('mode')
     if message_mode and message_mode in per_mode_send_queues:
+        # check for known corner case limitations and correct accordingly
+        if message['mode'] == 'call':
+            message_mode, message_body = check_call_restrictions(message)
+            message['mode'] = message_mode
+            message['body'] = message_body
```
dwang159: Are we overwriting the message here, or appending to the body? I'd prefer adding a blurb at the beginning/end of the message, rather than replacing the whole text.

src/iris/bin/sender.py
diegocepedaw: added

src/iris/bin/sender.py
diegocepedaw: The message body gets appended later so we're not overwriting the message, just prepending this warning. For extra robustness/futueproofness I added an extra check so that we append any existing message body.

src/iris/bin/sender.py
diegocepedaw: Fixed to only update message once. Also I pulled this functionality out into a helper function so other calling restrictions (like certain India calling restrictions) could be added in the future.

PR #521, opened 2019-05-18
Add unsubscribe page for apps supporting drop

dwang159:


src/iris/ui/static/js/iris.js
```diff
@@ -2803,7 +2808,59 @@ iris = {
       });
     }
   }, // End iris.appStats
-  singlestats:{},
```
diegocepedaw: I would verify that this doesn't break the singlestat endpoint. I think you'll have to add the singlestat base_route as an exception like you did unsubscribe.

PR #522, opened 2019-06-05
Add support for multi-recipient email notifications

dwang159:
This way, batch emails can be sent without hitting the API each
time. Instead, we send a single email with a list of recipients.
Note that this is currently supported only for out-of-band notifications
that specifically specify email as a contact mode, and does
not support priorities (where messages may be split across many
different contact modes).

### DIFF
--- a/src/iris/sender/rpc.py
+++ b/src/iris/sender/rpc.py
@@ -75,14 +75,15 @@ def handle_api_notification_request(socket, address, req):
         return
     notification['subject'] = '[%s] %s' % (notification['application'],
                                            notification.get('subject', ''))
+    target_list = notification.get('target_list')
     role = notification.get('role')
-    if not role:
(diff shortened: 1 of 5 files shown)


src/iris/sender/rpc.py
```diff
@@ -75,25 +75,44 @@ def handle_api_notification_request(socket, address, req):
...
     expanded_targets = None
     # if role is literal_target skip unrolling
     if not notification.get('unexpanded'):
-        try:
-            expanded_targets = cache.targets_for_role(role, target)
-        except IrisRoleLookupException:
-            expanded_targets = None
+        # For multi-recipient notifications, pre-populate destination with literal targets,
+        # then expand the remaining
+        if target_list:
```
diegocepedaw: might error incorrectly if there are only literal targets

PR #530, opened 2019-07-11
Custom sender address

diegocepedaw:
Add the ability to define an applications "from" address.


src/iris/ui/static/js/iris.js
```diff
@@ -2427,6 +2433,23 @@ iris = {
         self.modelPersist();
         self.render();
       });
+      data.$page.on('submit', data.addAddressForm, function(e) {
+        e.preventDefault();
+        var address = $('#add-address-box').val();
+        if (address == '') {
```
dwang159: Use '==='

src/iris/ui/static/js/iris.js
```diff
@@ -2427,6 +2433,23 @@ iris = {
...
+          iris.createAlert('Cannot add empty custom address');
+          return;
+        }
+        if(self.validateEmail(address) == false){
```
dwang159: ===

src/iris/api.py
```diff
@@ -2775,6 +2784,39 @@ def on_put(self, req, resp, app_name):
...
+                kill_address_modes = []
+
+                if not isinstance(new_addresses, dict):
+                    raise HTTPBadRequest('To change custom addresses, you must pass a dictionary of mode: address pairings')
+                for mode in new_addresses:
+                    if mode not in supported_custom_address_modes:
+                        raise HTTPBadRequest('%s does not support custom sender addresses', mode)
+
+                for mode in supported_custom_address_modes:
+                    if mode not in new_addresses:
```
dwang159: I would kill the address only if the caller specifies `email: None`. Otherwise, the caller needs to find out and then duplicate all the existing settings that they don't want to change.

src/iris/ui/templates/application.html
```diff
@@ -121,6 +121,24 @@ <h3>
...
 
+    {{#if @root.isInViewMode}}
+      <span class="light">Custom email "from" address: </span>
+      {{#if custom_sender_addresses.email}} 
+        <span class="label label-info">{{custom_sender_addresses.email}} <i class="glyphicon" data-variable="{{custom_sender_addresses.email}}"></i></span> 
+      {{else}}
+        <span class="light">None (using default sender address)</span>
+      {{/if}}
+    {{else}}
+      <span class="light">custom email "from" address:</span>
```
dwang159: Capitalize

src/iris/ui/templates/application.html
```diff
@@ -121,6 +121,24 @@ <h3>
...
+      {{#if custom_sender_addresses.email}}
+      <span class="label label-info">{{custom_sender_addresses.email}} <i class="glyphicon glyphicon-remove remove-address" data-variable="{{custom_sender_addresses.email}}"></i></span>
```
dwang159: This label/badge flow is confusing for an element that only takes a single value. We should switch to a text field more like the templates and less like the context variables. The top bar is also getting crowded, so we may want to move this and "supported modes" into a "Notification settings" subheader.

diegocepedaw: Changed UI to separate it out into its own div.

src/iris/api.py
diegocepedaw: I don't have strong feelings about it either way so I'll change it to the behaviour you mentioned.

PR #532, opened 2019-07-27
Category API

dwang159:


src/iris/api.py
```diff
@@ -4653,6 +4684,437 @@ def on_get(self, req, resp, incident_id):
         conn.close()
 
 
+class NotificationCategories(object):
+    allow_read_no_auth = True
+
+    def on_get(self, req, resp):
```
diegocepedaw: A use case we're probably going to employ in the front end is getting all the applications that have custom categories defined to populate the application dropdown/typeahead that users can add custom delivery preferences for. There probably isn't going to be a lot of applications with custom categories defined but it might be worth it to be able to return just a list of applications either through filtering or its own endpoint.

src/iris/api.py
```diff
@@ -4732,6 +5194,10 @@ def construct_falcon_api(debug, healthcheck_path, allowed_origins, iris_sender_a
     api.add_route('/v0/response/slack', ResponseSlack(iris_sender_app))
     api.add_route('/v0/twilio/deliveryupdate', TwilioDeliveryUpdate())
 
+    api.add_route('/v0/categories', NotificationCategories())
+    api.add_route('/v0/categories/{category_id}', NotificationCategory())
+    api.add_route('/v0/users/{username}/categories/{application}', CategoryOverrides())
```
diegocepedaw: It might be useful to add a /v0/users/{username}/categories endpoint to list all the overrides a user has defined together, or at least all the applications for which a user has defined overrides for. This would be useful for populating the user settings front end without having to make separate calls for every  application.

src/iris/api.py
dwang159: This can be handled via the existing Applications GET endpoint, with client-side filtering. We might want to implement server-side filters at some point though, as we get more applications. On the other hand, the number of apps is always going to be fairly limited.

PR #535, opened 2019-08-08
Notification types front end

diegocepedaw:
Implement front end for custom notification types and minor back end bug fixes.


src/iris/ui/static/js/iris.js
```diff
@@ -1890,16 +1891,23 @@ iris = {
       supportedTimezones: [],
       $page: $('.main'),
       $priority: $('#priority-table'),
+      $customNotificationTable: $('#custom-notification-table'),
```
dwang159: We should refer to user settings as "overrides" rather than "notifications". The current terminology is confusing, since we already use "notification" to refer to something else.

src/iris/ui/static/js/iris.js
```diff
@@ -2124,6 +2215,31 @@ iris = {
       self.data.appsToDelete[app] = true;
       self.createPriorityTable();
     },
+    deletePerCustomApp: function(elem) {
```
dwang159: This should be tied to a confirmation modal or something, to avoid misclicks.

src/iris/ui/static/js/iris.js
```diff
@@ -2213,6 +2356,45 @@ iris = {
...
     },
+    addApplicationNotification: function() {
+      var app = this.data.$addAppNotificationSelect.val();
+
+      this.data.$addAppNotificationSelect.val('');
+
+      if(this.data.settings.customApp.indexOf(app) === -1){
+        this.data.settings.customApp.push(app);
+      }
+      this.createCustomNotificationTable();
```
dwang159: The table entry for the app (on the main user settings page) appears before you click "Save" in the modal, and persists even if you cancel out of it. This entry should only appear after clicking "Save".

src/iris/ui/templates/application.html
```diff
@@ -302,6 +302,61 @@ <h4>Emails which create incidents</h4>
...
+    {{else}}
+      {{#if @root.isInViewMode}}
+        <p class="light">None configured.</p>
+      {{/if}}
+    {{/hasKeys}}
+    {{#ifNot @root.isInViewMode}}
+      {{#if @root.allowEditingCustomCategories}}
+        <form id="add-custom-category-form">
+          <input required type="text" maxlength="255" id="add-custom-category" class="form-control border-bottom" placeholder="Category name">
+          <input required type="text" maxlength="255" id="add-custom-category-description" class="form-control border-bottom " placeholder="Description">
```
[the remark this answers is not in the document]
diegocepedaw: I played around with it a bit and ended up enlarging it but keeping it an input. It looked more consistent than a textarea but still had quite a bit of room. That said my feelings on it are not super strong so I could still be open to changing it to a textarea if we coose to.

src/iris/ui/static/js/iris.js
```diff
@@ -1972,6 +1994,20 @@ iris = {
         iris.createAlert('Error: Failed to load reprioritization data -' + response.text);
       });
     },
+    getCustomNotificationSettings: function() {
+      var self = this;
+      return $.getJSON(this.data.url + this.data.user + '/categories').done(function(response){
+        self.data.settings.categories = response;
+        var tempSet = new Set();
```
jrgp: Instead of using `tempSet` as a variable name, why not something more descriptive like `applicationNames`?

src/iris/ui/static/js/iris.js
```diff
@@ -1972,6 +1994,20 @@ iris = {
...
+        self.data.settings.categories.forEach(function (item, index) {
+          tempSet.add(item['application']);
+        });
+        self.data.settings['customApp'] = Array.from(tempSet);
```
jrgp: As we're already using sets, why not just set self.data.settings.customApp to the set variable itself, instead of making it an array? 

That way you could use the native methods Set objects have (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set ) , instead of using splice/indexOf to check for presence/remove items.

src/iris/ui/static/js/iris.js
```diff
@@ -2111,6 +2147,61 @@ iris = {
...
+
+      // call out to the back end to get all categories for this app
+      var appNotificationCategories = [];
+      $.ajax({
+        type: "GET",
+        url: '/v0/categories/' + app,
+        success: function(resp)
+        {
+          appNotificationCategories = resp;
+          appNotificationCategories.forEach((appSetting, i) => {
```
jrgp: I'm seeing inconsistencies here as some parts/most of the Iris codebase uses ES5-style closures and this new code uses ES6 arrow functions.

I think a while ago Saif suggested all Iris code be ES5-compatible, as we're not using a transpiler like babel to convert new ES6 features like arrow functions to the ES5 equivalents.

On the other hand, all modern browsers support arrow functions, and if Iris used more of them the need for using variables named `self` for things like ajax callbacks would be obviated.

src/iris/ui/static/js/iris.js
```diff
@@ -2111,6 +2147,61 @@ iris = {
...
+              if(appSetting['application'] === userSetting['application'] && appSetting['name'] === userSetting['category']){
+                //apply user overrides
+                appNotificationCategories[i]['mode'] = userSetting['mode'];
+              }
+            });
+
+          });
+          settings['modalAppName'] = app;
+          settings['appCategoryData'] = appNotificationCategories;
+          settings['supported_modes'] = ['email','slack','call','sms','drop'];
```
jrgp: Would it be better if this hard coded list of modes be defined or derived from somewhere else in the codebase? 

With the list being hard coded here, if we ever need to add more modes, we'd need to remember to update this hard coded list. Also, `im` is missing.

src/iris/ui/static/js/iris.js
diegocepedaw: changed it to use set

diegocepedaw: changed to using sets

src/iris/ui/static/js/iris.js
diegocepedaw: I got rid of the arrow functions for consistency.

src/iris/ui/static/js/iris.js
diegocepedaw: changed

src/iris/ui/static/js/iris.js
```diff
@@ -2225,16 +2233,18 @@ iris = {
...
                     url: self.data.url + self.data.user + '/categories/' + app,
                     method: 'DELETE',
                     contentType: 'text'
                 }).done(function(){
-                    delete self.data.settings.customApp.splice(index, 1);
+                    // delete user mode overrides locally 
+                    self.data.settings.categories = self.data.settings.categories.filter(function( obj ) {
+                        return obj.application !== app;
+                    });
+                    delete self.data.settings.customApp.delete(app);
```
jrgp: Should this line have the `delete ` prefix?

src/iris/ui/static/js/iris.js
```diff
@@ -2321,15 +2331,22 @@ iris = {
                 contentType: 'text'
             }).done(function(){
                 iris.createAlert('Modal settings saved succesfully', 'success');
-                if (self.data.settings.customApp.indexOf(app) === -1) {
-                    self.data.settings.customApp.push(app);
+                if (!self.data.settings.customApp.has(app)) {
```
jrgp: the `! .has(app)` check before `.add(app)` isn't needed, as sets silently ignore dupes

src/iris/ui/static/js/iris.js
```diff
@@ -1990,6 +1993,14 @@ iris = {
         toggleReprioritization: function() {
             $('#reprioritization').toggleClass('active')
         },
+        getModes: function(){
```
jrgp: You might be able to use `window.appData.modes` to get at the list of modes, which would be a lot faster than adding another ajax call here.

src/iris/ui/static/js/iris.js
```diff
@@ -2225,16 +2233,18 @@ iris = {
         deletePerCustomApp: function(app) {
             var self = this;
 
-            var index = self.data.settings.customApp.indexOf(app);
-
-            if(index > -1){
-
+            if(self.data.settings.customApp.has(app)){
+                // delete user overrides for appliucation
```
jrgp: typo in `appliucation`

src/iris/ui/static/js/iris.js
```diff
@@ -2382,24 +2399,21 @@ iris = {
...
-                    Array.from(customOverrideApps).sort().forEach(function(app) {
-                        self.data.$addAppOverrideSelect.append($('<option>').text(app));
-                    });
-                }
-            }).fail(function(){
+                url: '/v0/categories/'
+            }).done(function (resp) {
+                  self.data.$addAppOverrideSelect.empty();
+                  self.data.$addAppOverrideSelect.append($('<option value="">').text('Add Application'));
+                  appOverrideCategories = resp;
```
jrgp: `appOverrideCategories` variable definition is missing the `var` prefix, so this variable becomes a global

src/iris/ui/static/js/iris.js
diegocepedaw: Good call, changed

PR #543, opened 2019-09-27
fix sync target team delete

diegocepedaw:
Prevent inactive teams from being pruned again and rename inactive teams to prevent namespace conflicts with new created teams.

### DIFF
--- a/src/iris/bin/sync_targets.py
+++ b/src/iris/bin/sync_targets.py
@@ -105,11 +105,14 @@ def prune_target(engine, target_name, target_type):
 
     if target_type == 'team':
         try:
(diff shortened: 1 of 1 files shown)


src/iris/bin/sync_targets.py
```diff
@@ -105,11 +105,14 @@ def prune_target(engine, target_name, target_type):
 
     if target_type == 'team':
         try:
-            engine.execute('''DELETE FROM `oncall_team` WHERE `target_id` = (SELECT `target`.`id` FROM `target` JOIN `target_type` ON `target`.`type_id` = `target_type`.`id` WHERE `target_type`.`name` = %s AND `target`.`name` = %s)''', target_type, target_name)
+            # rename team to prevent namespace conflict but preserve the ability to reactivate it in the future
+            new_name = str(uuid.uuid1())
```
jrgp: Might be better to use `uuid.uuid4()` instead, as that generates the UUID randomly rather than using a combination of the hostname + timestamp and is afaik more idiomatic.

PR #547, opened 2019-11-19
harden owa sync logic

colinxy:
- delete emails after successfully creating an incident
- batch updates/deletions
- bump exchangelib version

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -53,6 +53,7 @@ setuptools.setup(
         'prometheus': ['prometheus_client'],
         'dev': [
             'gunicorn',
+            'six',
             'Sphinx==1.5.6',
             'sphinxcontrib-httpdomain',
             'sphinx_rtd_theme',
--- a/setup.py
+++ b/setup.py
@@ -53,7 +53,6 @@ setuptools.setup(
         'prometheus': ['prometheus_client'],
         'dev': [
             'gunicorn',
-            'six',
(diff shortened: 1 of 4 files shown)


src/iris/bin/owasync.py
```diff
@@ -15,6 +15,10 @@
 import requests
 
 
+RELAY_IGNORE = 0
```
diegocepedaw: Is there a reason for differentiating between the RELAY_IGNORE and RELAY_SUCCESS? It seems like the code only checks if it was a relay failure or not and handles a relay_success and relay_ignore in the same way, perhaps we can avoid defining global variables and just return true or false for if the message was handled successfully or not.

src/iris/bin/run_server.py
```diff
@@ -8,7 +8,7 @@
 import sys
 import multiprocessing
 import gunicorn.app.base
-from gunicorn.six import iteritems
+from six import iteritems
```
diegocepedaw: The py3 branch already breaks python 2 compatibility. We can use items() instead and avoid importing the six library.

setup.py
```diff
@@ -53,6 +53,7 @@
         'prometheus': ['prometheus_client'],
         'dev': [
             'gunicorn',
+            'six',
```
jrgp: Should remove six as we aren't using it, as like Diego mentioned, this app is py3 only.

PR #548, opened 2019-11-20
add option to specify cert path directly

diegocepedaw:
Add the ability to explicitly set the ldap cert path for the sync_targets script.

### DIFF
--- a/src/iris/bin/sync_targets.py
+++ b/src/iris/bin/sync_targets.py
@@ -705,9 +705,8 @@ def main():
         # which will be in ldap already populated.
         if ldap_lists:
 
-            if 'ldap_cert_path' in config['auth']:
-                root = os.path.abspath('./')
-                ldap_cert_path = os.path.join(root, config['auth']['ldap_cert_path'])
+            if 'ldap_cert_path' in ldap_lists:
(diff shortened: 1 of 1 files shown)


src/iris/bin/sync_targets.py
```diff
@@ -700,6 +704,15 @@ def main():
...
         # which will be in ldap already populated.
         if ldap_lists:
+
+            if 'ldap_cert_path' in config['auth']:
+                root = os.path.abspath('./')
+                ldap_cert_path = os.path.join(root, config['auth']['ldap_cert_path'])
+                if not os.access(ldap_cert_path, os.R_OK):
+                    logger.error("Failed to read ldap_cert_path certificate")
+                    raise IOError
+                else:
```
jrgp: Do we need to try to force the path to be absolute using os.path.abspath and os.path.join? It's likely that the provided cert path in the config will be absolute, in which case the leading forward slash will make os.path.join() ignore the first argument. 

Also, instead pulling the cert path in by accessing the top-level `auth` subdict, can we instead add this `cert_path` as another key to the `ldap_lists` top level dict?

That way the configs for auth and ldap lists remain orthogonal.

PR #553, opened 2019-12-03
purge caches every sender loop and retry auditlog

diegocepedaw:
Because the targets cache is never purged if a team is renamed all plans that are now pointing to that new team name will fail because the targets cache will reflect the old name.

### DIFF
--- a/src/iris/sender/auditlog.py
+++ b/src/iris/sender/auditlog.py
@@ -2,6 +2,7 @@
 # See LICENSE in the project root for license information.
 
 from .. import db
+from gevent import sleep
 import logging
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


src/iris/sender/auditlog.py
```diff
@@ -15,11 +16,27 @@ def message_change(message_id, change_type, old, new, description):
...
-      VALUES (:message_id, :change_type, :old, :new, :description, NOW())
-    ''', dict(message_id=message_id, change_type=change_type, old=old, new=new, description=description))
-    session.commit()
-    session.close()
-    logger.info('Logged change information for message (ID %s)', message_id)
+    # retry to guard against deadlocks
+    retries = 0
+    max_retries = 5
+    while True:
+        with db.guarded_session() as session:
```
jrgp: We shouldn't use both this `db.guarded_session()` context manager as well as `db.Session()` on line 26 to instantiate a session, as only the session created on line 26 is getting used.

diegocepedaw: fixed

PR #560, opened 2020-01-23
Add restricted country sms override

diegocepedaw:
To ensure successful SMS delivery in countries with restrictions like India messages must adhere to a pre-registered template. This PR allows users to override SMS bodies with the whitelisted template to prevent carriers from dropping their messages.

### DIFF
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -3788,9 +3788,7 @@ class User(object):
                              JOIN `mode` ON `mode`.`id` = `mode_template_override`.`mode_id`
                          WHERE `mode_template_override`.`target_id` = %s'''
         cursor.execute(mode_template_override_query, user_id)
-        user_data['template_overrides'] = []
-        for row in cursor:
-            user_data['template_overrides'].append(row['mode'])
+        user_data['template_overrides'] = [row['mode'] for row in cursor]
 
         # Get user contact modes
         modes_query = '''SELECT `priority`.`name` AS priority, `mode`.`name` AS `mode`
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -525,6 +525,20 @@ WHERE `target_id` = (SELECT `id` from `target` WHERE `name` = :name AND `type_id
       `priority_id` = (SELECT `id` from `priority` WHERE `name` = :priority) AND
       `application_id` = (SELECT `id` from `application` WHERE `name` = :app)'''
 
+insert_user_modes_template_override_query = '''INSERT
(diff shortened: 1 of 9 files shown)


src/iris/api.py
```diff
@@ -3768,6 +3782,16 @@ def on_get(self, req, resp, username):
...
 
+        # get any mode based template override settings
+        mode_template_override_query = '''SELECT `mode`.`name` AS `mode`
+                         FROM `mode_template_override`
+                             JOIN `mode` ON `mode`.`id` = `mode_template_override`.`mode_id`
+                         WHERE `mode_template_override`.`target_id` = %s'''
+        cursor.execute(mode_template_override_query, user_id)
+        user_data['template_overrides'] = []
+        for row in cursor:
+            user_data['template_overrides'].append(row['mode'])
```
jwon: You can use list comprehension!

```
user_data['template_overrides'] = [row['mode'] for row in cursor]
```

src/iris/api.py
```diff
@@ -3900,6 +3922,52 @@ def on_put(self, req, resp, username):
...
+    enforce_user = True
+
+    def on_post(self, req, resp, username):
+        # check request body integrity
+        try:
+            data = ujson.loads(req.context['body'])
+        except ValueError:
+            raise HTTPBadRequest('Invalid json in post body')
+        if not isinstance(data.get('template_overrides'), dict):
+            raise HTTPBadRequest('Invalid json in post body')
```
jwon: Use a unique error string for each error so that it's easily identifiable where in the code the exception occurred. 

I think this one can be something like `"Invalid json in post body: 'template_overrides' is not a dict"` or something like that

src/iris/api.py
```diff
@@ -3900,6 +3922,52 @@ def on_put(self, req, resp, username):
...
+        elif override_val == 'disabled':
+            try:
+                cursor.execute(delete_user_modes_template_override_query, req.context['username'])
+                connection.commit()
+            except Exception:
+                logger.exception('Failed setting mode template override for user %s', req.context['username'])
+        else:
+            cursor.close()
+            connection.close()
+            raise HTTPBadRequest('Invalid sms override setting value')
```
jwon: Maybe also print what the value was? 

`raise HTTPBadRequest(f'Invalid sms override setting value: {override_val}')`

src/iris/api.py
diegocepedaw: sure, changed

src/iris/api.py
diegocepedaw: sure, changed

PR #561, opened 2020-01-30
fix slack error handling and task failure counts

diegocepedaw:


src/iris/bin/sender.py
```diff
@@ -1298,23 +1298,22 @@ def fetch_and_send_message(send_queue, vendor_manager):
...
-            try:
-                success, sent_locally = distributed_send_message(message, vendor_manager)
-            # nope - log and bail
-            except Exception:
-                metrics.incr('task_failure')
-                add_mode_stat(message['mode'], None)
-                logger.error('unable to send %(mode)s %(message_id)s %(application)s %(destination)s %(subject)s %(body)s', message)
-                sent_locally = True
+        success = False
+        sent_locally = False
```
colinxy: This is not necessary since we already set them before try/except block.

PR #564, opened 2020-02-13
Make noisy "Relaying message to current master sender" log message DEBUG

jwon:
Let's make the logs more easily consumable!

### DIFF
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -2260,7 +2260,7 @@ class Notifications(object):
             with Timeout(self.timeout, False):
                 sender_addr = self.coordinator.get_current_master()
             if sender_addr:
-                logger.info('Relaying message to current master sender: %s', sender_addr)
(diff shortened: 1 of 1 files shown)


diegocepedaw:
While you are at it do you want to also change the loglevel of the Incidents() on_post method insertion retry to a warning so we don't have a bunch of stack traces in the logs `logger.exception('Failed inserting incident. (Try %s/%s)', retries, max_retries)` retries to warning

jwon:
I disagree, I think we should keep the exception chain because we could be retrying for different types of exceptions. We would want to know that. 

I think right now the overly verbose exception chain is indeed annoying but we should fix it by fixing the root cause, not by reducing logging.

diegocepedaw:
We could keep the exception logging on breaching the retries to bubble up any non-transient issues while not filling up the logs with long tracebacks that can potentially make it harder to notice other issues. I do see the merit in your argument though so I'll leave it up to you. This change looks good to me.

PR #565, opened 2020-02-14
Add db check to healthcheck endpoint

diegocepedaw:
Return a 503 from the healthcheck endpoint if the database is unreachable;


diegocepedaw:
need to update tests

test/test_irisapi.py
```diff
@@ -35,8 +35,8 @@ def test_healthcheck(self):
             self.api.add_route('/healthcheck', Healthcheck('healthcheck_path'))
             result = self.simulate_get(path='/healthcheck')
             m.assert_called_once_with('healthcheck_path')
-        self.assertEqual(result.status_code, 200)
-        self.assertEqual(result.content, b'GOOD')
+        self.assertEqual(result.status_code, 503)
```
diegocepedaw: This will test for failure because the db is not set up for this test and the e2e test will test for success because there the db is set up

PR #570, opened 2020-03-20
add distributed incident creation rate limiting

diegocepedaw:
if a plan breaches the quota set by incident_quota_max
        it will no longer be able to create incidents. After
        quota_breach_duration amount of seconds from the first incident
        the plan's quota will reset.

### DIFF
--- a/src/iris/cache.py
+++ b/src/iris/cache.py
@@ -119,14 +119,11 @@ def check_quota_breach(plan):
     # plans can have spaces and keys can't, hash to normalize keys
     plan_hash = str(hash(plan))
     if not mc.get(plan_hash):
-        print(quota_configs.get('quota_breach_duration'))
         # set expiry to quota_breach_duration seconds, after expiry quota resets
         mc.set(plan_hash, 0, quota_configs.get('quota_breach_duration'))
     # atomic memcached-side increment operation
     mc.incr(plan_hash)
     quota_used = mc.get(plan_hash)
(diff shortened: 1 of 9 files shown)


src/iris/cache.py
```diff
@@ -92,7 +97,58 @@ def cache_modes():
...
+        it will no longer be able to create incidents. After
+        quota_breach_duration amount of seconds from the first incident
+        the plan's quota will reset.
+    '''
+    global quota_configs
+    if quota_configs.get('disable_quota'):
+        # skip quota check
+        return False
+    try:
+        mc = pylibmc.Client([quota_configs.get('memcache_host')], binary=True, behaviors={"cas": True, "tcp_nodelay": True, "ketama": True})
```
colinxy: You are creating a new memcached client every time `check_quota_breach` is called. This code is in the hot path of incident creation, so it's probably a good idea to use connection pooling.

src/iris/cache.py
```diff
@@ -92,7 +97,55 @@ def cache_modes():
...
```
colinxy: You are creating a new memcached client every time check_quota_breach is called. This code is in the hot path of incident creation, so it's probably a good idea to use connection pooling.

diegocepedaw: Good point, I'll add that

diegocepedaw:
repaced with NGINX rate rimiting

PR #572, opened 2020-03-23
only log incident tracebacks when retries breached

diegocepedaw:
For incident insertion sqlalchey InternalErrors and OperationalErrors only log the stack trace if retries are breached. otherwise,  simply log an error.

### DIFF
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -1717,7 +1717,7 @@ class Incidents(object):
 
                     session.commit()
                     session.close()
-                except Exception as e:
+                except (InternalError, OperationalError) as e:
                     logger.error('Failed inserting incident for plan %s. (Try %s/%s)', plan_id, retries, max_retries)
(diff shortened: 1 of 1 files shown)


src/iris/api.py
```diff
@@ -1718,13 +1718,13 @@ def on_post(self, req, resp):
                     session.commit()
                     session.close()
                 except (InternalError, OperationalError) as e:
-                    logger.exception('Failed inserting incident. (Try %s/%s)', retries, max_retries)
+                    logger.error('Failed inserting incident. (Try %s/%s)', retries, max_retries)
```
jwon: Since we're only logging the error message now (no stack trace), we should also log the `plan_id` or something additional to give context as to what incident we're talking about when we see this log message in the logs.

diegocepedaw: sure

PR #574, opened 2020-04-02
ldap sync catch exceptions

diegocepedaw:


src/iris/bin/sync_targets.py
```diff
@@ -668,7 +668,7 @@ def main():
     config = load_config()
     metrics.init(config, 'iris-sync-targets', stats_reset)
 
-    default_ldap_timeout = 20
+    default_ldap_timeout = 60
     default_nap_time = 3600
 
     ldap_timeout = int(config.get('sync_script_ldap_timeout', default_ldap_timeout))
```
colinxy, line 674: Can you add this to the example config also? Ideally we would like to easily bump it when we get an incident

PR #585, opened 2020-08-12
target_sync improvements

diegocepedaw:
- target_sync bug fixes
- remove batch message insert
- commit user's new mailing list memberships individually instead of all together 

Using a synchronously replicated database like Galera it is best to avoid large transactions as they impact replication performance. Due to this, it is actually faster and less prone to failures to split batched transactions up especially when write loads are high like with Iris.

### DIFF
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -4041,8 +4041,19 @@ class ResponseMixin(object):
                 'body': body,
                 'destination': dest
             }
-            message_id = session.execute(sql, data).lastrowid
-            session.commit()
+            retries = 0
+            while retries < 3:
+                try:
+                    message_id = session.execute(sql, data).lastrowid
+                    session.commit()
+                except Exception as e:
+                    retries += 1
+                    if retries == 3:
+                        logger.exception("failed to create email message")
+                    sleep(0.2)
+                else:
+                    break
+
             session.close()
             return True, message_id
 
@@ -4240,7 +4251,7 @@ class ResponseEmail(ResponseMixin):
         try:
             app, response = self.handle_user_response('email', msg_id, source, cmd)
         except Exception:
-            logger.exception('Failed to handle email response: %s' % first_line)
+            logger.warning('Failed to handle email response: %s' % first_line)
             raise
 
         # When processing a claim all scenario, the first item returned by handle_user_response
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -4041,18 +4041,8 @@ class ResponseMixin(object):
                 'body': body,
                 'destination': dest
             }
-            retries = 0
-            while retries < 3:
-                try:
-                    message_id = session.execute(sql, data).lastrowid
-                    session.commit()
-                except Exception as e:
-                    retries += 1
-                    if retries == 3:
-                        logger.exception("failed to create email message")
-                    sleep(0.2)
-                else:
-                    break
+            message_id = session.execute(sql, data).lastrowid
+            session.commit()
 
             session.close()
             return True, message_id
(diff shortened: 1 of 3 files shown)


src/iris/api.py
```diff
@@ -4041,8 +4041,19 @@ def create_email_message(self, application, dest, subject, body):
                 'body': body,
                 'destination': dest
             }
-            message_id = session.execute(sql, data).lastrowid
-            session.commit()
+            retries = 0
+            while retries < 3:
```
sp1ff, line 4045: Can you say a bit about why you're doing this? Do we know this statement to be failing? If so, why? Can we address the underlying problem? Re-trying 3 times with a sleep of 0.2 sec in between seems arbitrary. If you have data to support these values, well & good: document them.
It also seems like the sort of thing for which we should be accumulating metrics, since we'll presumably want to monitor the situation.

src/iris/bin/sync_targets.py
```diff
@@ -391,8 +391,7 @@ def sync_from_oncall(config, engine, purge_old_users=True):
             except SQLAlchemyError as e:
                 logger.exception('Error inserting oncall_team %s: %s', t, e)
                 continue
-
-    session.commit()
+        session.commit()
```
sp1ff, line 394: Whoah-- this function has ~400 LoC. That needs to be addressed before doing anything else. I defy you to explain to me the impact of this change for every code path through this method.

src/iris/api.py
```diff
@@ -4041,8 +4041,19 @@ def create_email_message(self, application, dest, subject, body):
...
+                try:
+                    message_id = session.execute(sql, data).lastrowid
+                    session.commit()
+                except Exception as e:
+                    retries += 1
+                    if retries == 3:
+                        logger.exception("failed to create email message")
+                    sleep(0.2)
```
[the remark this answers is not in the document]
jwon: I think this is fine, as `session.commit()` will probably emit a nested exception and logging that should give us context as to what went wrong. 

My only feedback here is maybe we should log the sql statement it was trying to execute when the exception occurred?

src/iris/api.py
```diff
@@ -4240,7 +4251,7 @@ def on_post(self, req, resp):
         try:
             app, response = self.handle_user_response('email', msg_id, source, cmd)
         except Exception:
-            logger.exception('Failed to handle email response: %s' % first_line)
+            logger.warning('Failed to handle email response: %s' % first_line)
```
jwon, line 4254: Why this change? Don't we want to log the exception?

src/iris/api.py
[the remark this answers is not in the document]
sp1ff: `session.commit` is in the try block. It will arrive as `e`. We should capture the information in that exception, not just log a messages saying "failed to create..."

src/iris/api.py
diegocepedaw: This was causing a ton of noise in the logs due to users responding with an incorrect message id. Since this can fail in several ways from user input and is not critical I think logging an exception is overkill.

src/iris/api.py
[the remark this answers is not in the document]
diegocepedaw: The sqlalchemy exception information that is logged along with that message will contain the query and values that were tried. The point is moot since I am removing this message retry because after reconsidering since retrying a message that confirms an incident was claimed via email does not seem pressing enough especially if it is underlying database issues that caused the failure, to begin with.

src/iris/bin/sync_targets.py
diegocepedaw: Reverted since this would be unnecessary as the inserts in this function directly use engine.execute() so the session doesn't actually have anything to commit. I completely agree that this should be refactored into smaller functions though perhaps that is best left to a different pull request. I think taking a second look at how we use sqlalchemy, if not replace it entirely, is in the horizon but again a bit out of scope for this change.

PR #594, opened 2020-10-13
fix tracking bug

diegocepedaw:


src/iris/bin/sender.py
```diff
@@ -1663,6 +1674,7 @@ def init_sender(config):
     db.init(config)
     cache.init(api_host, config)
     metrics.init(config, 'iris-sender', default_sender_metrics)
+    spawn(metrics.emit_forever)
```
colinxy, line 1677: This will likely give us a lot of stale metrics if a sync loop does not finish within a minute. I would prefer if we can at least send zero values than old values from the previous loop.

I see that you added metric `last_completed_sender_loop` for us to figure out when last loop finished, but we now need to compare any metric with `last_completed_sender_loop` side by side to determine if a value is stale.

diegocepedaw: Staleness wouldn't be more of an issue than it currently is since we have default values for these metrics and after emitting the current values they are reset to the default values (which is just zeroes) so they would actually emit zeroed values for the metrics. However, I am reverting to emitting at the end of the loop since there is an edge case where if the loops goes barely past the minute it could complete another loop before metrics are emitted again potentially overwriting some metric values. This could be mitigated by aggregating the metrics from those two runs but I think the better solution is to enforce the ssender loop staying under a minute.

colinxy (approved):

PR #600, opened 2021-02-03
Add enpoint to get user's slack id

diegocepedaw:
In order to tag a user in a slack message created via an api request, it is necessary to use their slack id. This endpoint will allow clients to resolve that slack id from a username so it can be incorporated in the body of an iris message.

### DIFF
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -33,6 +33,7 @@ from . import cache
 from . import ui
 from . import app_stats
 from .config import load_config
+from iris.vendors.iris_slack import iris_slack
 from iris.sender import auditlog
 from iris.sender.quota import (get_application_quotas_query, insert_application_quota_query,
                                required_quota_keys, quota_int_keys)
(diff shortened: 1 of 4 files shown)


minhaz1:
Might be a good idea to cache the results for this. The username -> slack id mapping shouldn't change. Could safely put in a cache with a pretty long TTL. Could also add caching at the nginx tier.

src/iris/api.py
```diff
@@ -4986,6 +4987,65 @@ def on_post(self, req, resp, application):
...
+        cursor.execute(query, query_params)
+        user_email = cursor.fetchone()
+        cursor.close()
+        conn.close()
+        if user_email is None:
+            raise HTTPBadRequest('User does not exist in Iris')
+
+        # query slack for user id from email address
+        slack_vendor = iris_slack(self.slack_cfg)
+        slack_id = slack_vendor.lookup_by_email(user_email['destination'])
```
colinxy, line 5050: `lookup_by_email` can raise an exception

src/iris/vendors/iris_slack.py
```diff
@@ -32,6 +32,26 @@ def __init__(self, config):
...
+                                     proxies=self.proxy,
+                                     timeout=self.timeout)
+            if response.status_code == 200:
+                data = response.json()
+                if data.get('ok') and data.get('user'):
+                    if data['user'].get('id'):
+                        return data['user']['id']
+
+            logger.error('Failed resolve user id from email: %d', response.status_code)
+            return False
```
colinxy, line 50: I think it's better to surface the upstream error message to our users than returning false

colinxy (changes requested):

src/iris/api.py
diegocepedaw: changed it so that it's caught and instead raises HTTPInternalServerError() in the api

src/iris/vendors/iris_slack.py
diegocepedaw: The user will get a 404 with a 'Failed to get id from Slack' message, not sure if additional specificity will be more useful to the end-user

src/iris/api.py
```diff
@@ -5043,17 +5044,22 @@ def on_get(self, req, resp, username):
...
+            resp.body = 'User does not exist in Iris'
+            raise HTTPBadRequest()
 
         # query slack for user id from email address
         slack_vendor = iris_slack(self.slack_cfg)
-        slack_id = slack_vendor.lookup_by_email(user_email['destination'])
+        try:
+            slack_id = slack_vendor.lookup_by_email(user_email['destination'])
+        except Exception as e:
+            raise HTTPInternalServerError()
```
colinxy, line 5055: Can you also send the exception detail as part of the response?

colinxy (changes requested):

src/iris/api.py
diegocepedaw: sure

colinxy (approved):

PR #602, opened 2021-02-08
 create a new resolved attribute for incidents [DB SCHEMA CHANGE]

diegocepedaw:
This new "resolved" attribute allows the distinction between an incident that has been claimed but is still an ongoing issue and an incident whose cause has been mitigated.


jrgp (approved):
LGTM

PR #603, opened 2021-02-25
Add a configurable default per app quota for message creation

diegocepedaw:
previously the default behaviour was no message creation quotas unless a user specified one on the application page. Now there is a configurable default quota that will apply to all application without a defined quota. If a quota is specified in the application page that will supersede the default quota settings.

### DIFF
--- a/configs/config.dev.yaml
+++ b/configs/config.dev.yaml
@@ -135,6 +135,25 @@ sender:
   ## slaves to be cycled through
   is_master: True
 
+  # default_rate_def:
+  #   # hard quota 60 messages in a period, soft quota 30 messages in a period, 1 minute HQ period length, 1 minute SQ period length, no plans/targets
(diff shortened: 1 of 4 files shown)


configs/config.dev.yaml
```diff
@@ -135,6 +135,25 @@ sender:
   ## slaves to be cycled through
   is_master: True
 
+  # default_rate_def:
+  #   # hard quota 60 messages in a period, soft quota 30 messages in a period, 1 minute HQ period length, 1 minute SQ period length, no plans/targets
+  #   #self.default_rate_def = (40,30,1,1,3600,None,None)
```
jwon, line 140: should this be 60 not 40 to be consistent with comment above?

jwon (approved):

configs/config.dev.yaml
diegocepedaw: that was a comment I was using for reference and missed taking out, removed

PR #606, opened 2021-03-18
 add customizable post incident creation handler

diegocepedaw:
Add optional incident handler to do additional tasks after the incident has been created

### DIFF
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -1694,12 +1694,6 @@ class Incidents(object):
             if not app_template_count:
                 raise HTTPBadRequest('No plan template actions exist for this app')
 
-        incident_data = {
-            'plan': incident_params['plan'],
-            'created': int(time.time()),
-            'application': req.context['app']['name'],
-            'context': context
-        }
 
         # To try to avoid deadlocks, split the inserts into their own session
         retries = 0
@@ -1752,6 +1746,15 @@ class Incidents(object):
         resp.status = HTTP_201
         resp.set_header('Location', '/incidents/%s' % incident_id)
         resp.body = ujson.dumps(incident_id)
+
+
+        incident_data = {
+            'id': incident_id,
+            'plan': incident_params['plan'],
+            'created': int(time.time()),
+            'application': req.context['app']['name'],
+            'context': context
+        }
 
         # optional incident handler to do additional tasks after the incident has been created
         if self.custom_incident_handler_module is not None:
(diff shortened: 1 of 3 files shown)


src/iris/api.py
```diff
@@ -1737,6 +1746,15 @@ def on_post(self, req, resp):
         resp.set_header('Location', '/incidents/%s' % incident_id)
         resp.body = ujson.dumps(incident_id)
 
+        # optional incident handler to do additional tasks after the incident has been created
+        if self.custom_incident_handler_module is not None:
+            connection = db.engine.raw_connection()
+            cursor = connection.cursor(db.dict_cursor)
+            cursor.execute(single_incident_query, int(incident_id))
```
colinxy, line 1753: You already have the data for the incident above, so this query shouldn't be necessary

diegocepedaw: In this case additional sql queries will be required to fetch plan and plan step info

src/iris/api.py
```diff
@@ -1685,6 +1694,13 @@ def on_post(self, req, resp):
             if not app_template_count:
                 raise HTTPBadRequest('No plan template actions exist for this app')
 
+        incident_data = {
```
colinxy, line 1697: Are you sure you don't want to include incident_id as part of this data so that people can link them on iris as well?

src/iris/api.py
```diff
@@ -1737,6 +1753,33 @@ def on_post(self, req, resp):
         resp.set_header('Location', '/incidents/%s' % incident_id)
         resp.body = ujson.dumps(incident_id)
 
+        # optional incident handler to do additional tasks after the incident has been created
+        if self.custom_incident_handler_module is not None:
+            connection = db.engine.raw_connection()
+            cursor = connection.cursor(db.dict_cursor)
+
+            # get plan info
```
colinxy, line 1761: this section smells like the same logic for query a plan. Can you abstract them away in another function?

src/iris/api.py
diegocepedaw: good catch, added that

src/iris/api.py
diegocepedaw: It's the same code but I don't think it's worth abstracting away since it's a pretty small snippet that isn't going to get reused anywhere else, and the end result would be slightly less readable. I just doesn't seem necessary especially since all the endpoints currently handle everything within their own class anyways.

colinxy (approved):

PR #608, opened 2021-04-21
add internal only application endpoint

diegocepedaw:
- adds internal API endpoint to return all application auth details for use by a custom sender that does not have direct access to the database
- add a category override and device registration information to user endpoint

### DIFF
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -5407,7 +5407,7 @@ class CategoryOverrides(object):
         resp.status = HTTP_204
 
 
-class InternalApplicationsAuth(object):
+class InternalApplicationsAuth():
     allow_read_no_auth = False
 
     def __init__(self, config):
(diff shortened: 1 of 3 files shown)


src/iris/api.py
```diff
@@ -5388,6 +5388,29 @@ def on_delete(self, req, resp, username, application):
         resp.status = HTTP_204
 
 
+class InternalApplicationsAuth(object):
```
jrgp, line 5391: Since we moved off of python 2.7, we don't need to inherit from object anymore.

jrgp (changes requested):

src/iris/api.py
diegocepedaw: good point, we should do a cleanup of any leftover python 2.7 cruft

jrgp (approved):

PR #612, opened 2021-04-28
internal target resolution and incident escalation

diegocepedaw:
- endpoint to take role:target pair and return list of user details
- endpoint to increment incident steps and mark them active/inactive

### DIFF
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -3976,7 +3976,7 @@ def get_user_details(username):
                         FROM `device`
                         WHERE `device`.`user_id` = %s'''
     cursor.execute(mode_template_override_query, user_id)
-    user_data['device'] = [row for row in cursor]
+    user_data['device'] = list(cursor)
 
     # Get user contact modes per app
     user_data['per_app_modes'] = defaultdict(dict)
@@ -3990,8 +3990,7 @@ def get_user_details(username):
                     WHERE `user_team`.`user_id` = %s'''
     cursor.execute(teams_query, user_id)
     user_data['teams'] = []
-    for row in cursor:
-        user_data['teams'].append(row['team'])
+    user_data['teams'] = [row['team'] for row in cursor]
 
     # Get user contact info
     contacts_query = '''SELECT `mode`.`name` AS `mode`, `target_contact`.`destination` AS `destination`
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -3956,7 +3956,7 @@ def get_user_details(username):
                                     JOIN `application` ON `application`.`id` = `notification_category`.`application_id`
                                     WHERE `category_override`.`user_id` = %s'''
     cursor.execute(category_override_query, user_id)
-    user_data['category_overrides'] = defaultdict(dict)
+    user_data['category_overrides'] = {}
     for row in cursor:
         user_data['category_overrides'][row['application_name']] = row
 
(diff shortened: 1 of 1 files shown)


src/iris/api.py
```diff
@@ -5430,10 +5438,82 @@ def on_get(self, req, resp):
...
+
+
+class InternalTarget():
+    allow_read_no_auth = False
+
+    def __init__(self, config):
+        self.allowlisted_apps = config.get('allowlisted_internal_apps', [])
+
+    def on_get(self, req, resp):
+        if req.context.get('app', {}).get('name') not in self.allowlisted_apps:
```
jrgp, line 5487: In the future, it might be a good idea to use a middleware or a decorator for this copy/pasted auth logic.

src/iris/api.py
```diff
@@ -5430,10 +5438,82 @@ def on_get(self, req, resp):
...
+            raise HTTPBadRequest('Missing required fields "current_step" or "active" in body')
+
+        if not isinstance(body['current_step'], int):
+            raise HTTPBadRequest('"step" field must be an integer"')
+
+        connection = db.engine.raw_connection()
+        cursor = connection.cursor(db.dict_cursor)
+
+        cursor.execute(''' SELECT EXISTS (SELECT `id` from `incident` where `id` = %s)''', incident_id)
+        if cursor.fetchone() is None:
```
jrgp, line 5463: Because this is doing a ` select exists()` won't the return value always be a single-element array containing a 1 or 0 as opposed to something that could be none?

src/iris/api.py
```diff
@@ -3920,91 +3922,97 @@ def on_get(self, req, resp):
...
+
+    # get any category override settings
+    category_override_query = '''SELECT `application`.`name` AS `application_name`, `mode`.`name` AS `mode`, `notification_category`.`name` AS `category`
+                                    FROM `category_override`
+                                    JOIN `mode` ON `mode`.`id` = `category_override`.`mode_id`
+                                    JOIN `notification_category` ON `notification_category`.`id` = `category_override`.`category_id`
+                                    JOIN `application` ON `application`.`id` = `notification_category`.`application_id`
+                                    WHERE `category_override`.`user_id` = %s'''
+    cursor.execute(category_override_query, user_id)
+    user_data['category_overrides'] = defaultdict(dict)
```
jrgp, line 3959: Does this need to be a default dict?

jrgp (changes requested):

src/iris/api.py
diegocepedaw: good idea, I'll include that when I add the next internal endpoints

src/iris/api.py
```diff
@@ -3920,91 +3922,97 @@ def on_get(self, req, resp):
...
+    user_data['modes'] = {}
+    for row in cursor:
+        user_data['modes'][row['priority']] = row['mode']
+
+    # get device settings for user
+    mode_template_override_query = '''SELECT `device`.`registration_id`, `device`.`user_id`, `device`.`platform`
+                        FROM `device`
+                        WHERE `device`.`user_id` = %s'''
+    cursor.execute(mode_template_override_query, user_id)
+    user_data['device'] = [row for row in cursor]
```
jrgp, line 3979: Can this list comp be simplified to `list(cursor)` ?

src/iris/api.py
```diff
@@ -3920,91 +3922,97 @@ def on_get(self, req, resp):
...
+        user_data['per_app_modes'][row['application']][row['priority']] = row['mode']
+
+    # Get user teams
+    teams_query = '''SELECT `target`.`name` AS `team`
+                    FROM `user_team` JOIN `target` ON `user_team`.`team_id` = `target`.`id`
+                    WHERE `user_team`.`user_id` = %s'''
+    cursor.execute(teams_query, user_id)
+    user_data['teams'] = []
+    for row in cursor:
+        user_data['teams'].append(row['team'])
```
jrgp, line 3994: Like on 3949, can this be simplified to the following?

    user_data['teams'] = [row['team'] for row in cursor]

jrgp (changes requested):

jrgp (approved):

PR #621, opened 2021-06-16
skip error logging if iris bot fails to send messages to a channel it is not in

ddurruty-li:
skip error logging if iris bot fails to send messages to a channel it is not in

### DIFF
--- a/src/iris/vendors/iris_slack.py
+++ b/src/iris/vendors/iris_slack.py
@@ -126,15 +126,9 @@ class iris_slack(object):
                     return time.time() - start
                 # If message is invalid:
                 #   {u'ok': False, u'error': u'invalid_arg_name'}
-
-                # Check the slack api to see if the iris bot is in the channel first before logging an error
-                # if not in the channel, the error is expected and should not be logged
-                msg_endpoint = self.config['base_url'] + "/conversations.info"
-                response_alt = requests.get(msg_endpoint, data=payload, proxies=self.proxy, timeout=self.timeout)
-                if response_alt.status_code == 200 \
-                        and response_alt.json()['ok'] \
-                        and not response_alt.json()['channel']['is_member']:
-                    pass
+                # if not in the channel, the error is expected so log a warning instead of an error
+                elif data.get('error') == 'not_in_channel':
+                    logger.warning('Iris bot not present in the designated channel %s', message.get('destination'))
                 else:
                     logger.error('Received an error from slack api: %s', data['error'])
             elif response.status_code == 429:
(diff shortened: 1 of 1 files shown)


src/iris/vendors/iris_slack.py
```diff
@@ -125,8 +126,17 @@ def send_message(self, message):
...
+
+                # Check the slack api to see if the iris bot is in the channel first before logging an error
+                # if not in the channel, the error is expected and should not be logged
+                msg_endpoint = self.config['base_url'] + "/conversations.info"
+                response_alt = requests.get(msg_endpoint, data=payload, proxies=self.proxy, timeout=self.timeout)
+                if response_alt.status_code == 200 \
+                        and response_alt.json()['ok'] \
+                        and not response_alt.json()['channel']['is_member']:
+                    pass
+                else:
```
diegocepedaw, line 138: we probably still want to log at least an Info line so that we can easily see if there are users who are missing message deliveries

src/iris/vendors/iris_slack.py
```diff
@@ -125,8 +126,17 @@ def send_message(self, message):
                     return time.time() - start
                 # If message is invalid:
                 #   {u'ok': False, u'error': u'invalid_arg_name'}
-                logger.error('Received an error from slack api: %s',
-                             data['error'])
+
+                # Check the slack api to see if the iris bot is in the channel first before logging an error
+                # if not in the channel, the error is expected and should not be logged
+                msg_endpoint = self.config['base_url'] + "/conversations.info"
```
diegocepedaw, line 132: this api call is unnecessary, slack api already returns a `not_in_channel error`, you should be able to the same thing we do above in line 123

diegocepedaw (changes requested):

src/iris/vendors/iris_slack.py
```diff
@@ -125,8 +126,11 @@ def send_message(self, message):
                     return time.time() - start
                 # If message is invalid:
                 #   {u'ok': False, u'error': u'invalid_arg_name'}
-                logger.error('Received an error from slack api: %s',
-                             data['error'])
+                # if not in the channel, the error is expected so log a warning instead of an error
+                elif data.get('error') == 'not_in_channel':
+                    logger.warning('Iris bot not present in the designated channel %s', message.get('destination'))
```
diegocepedaw, line 131: I believe you have to return a runtime `return time.time() - start` or else it will still be counted in the failure metrics

diegocepedaw (changes requested):

PR #627, opened 2021-07-29
added twillio number override mechanism

ddurruty-li:


src/iris/vendors/iris_twilio.py
```diff
@@ -78,7 +78,10 @@ def send_sms(self, message):
         client = self.get_twilio_client()
 
         sender = client.messages.create
-        from_ = self.config['twilio_number']
+        if message['application'] in self.config['application_override_mapping']:
```
diegocepedaw, line 81: Does this work correctly when the `application_override_mapping` config is not defined? Might potentially cause a keyerror exception if Twilio vendor is enabled but application_override_mapping is left commented out. Maybe fetch with self.config.get('application_override_mapping', {}) instead.

diegocepedaw (changes requested):

src/iris/vendors/iris_twilio.py
ddurruty-li: Updated commit and removed the small change I had made to two tests I had made since it's no longer necessary

PR #632, opened 2021-08-12
Iris-message-processor cluster management

diegocepedaw:
- all incidents are randomly assigned a bucket id
- when sender nodes check in they are randomly assigned a slice of buckets to process
- if a sender node does not check in for longer than the ttl their buckets are reassigned to the rest of the cluster
- if more sender nodes join existing buckets are redistributed equally
- add internal incident endpoint to fetch incidents that correspond to the bucket assignments of a sender node, this is split into a get that returns a list of ids and a post that takes a list of ids and returns details so that the sender client can do batching if necessary


src/iris/utils.py
```diff
@@ -9,17 +9,29 @@
...
 import re
 import msgpack
 import logging
+from random import randrange
 
 logger = logging.getLogger(__name__)
 
+config = {}
+if 'IRISPYTEST' not in os.environ:
+    config = load_config()
```
jwon, line 24: why is this necessary?

src/iris/api.py
```diff
@@ -5604,6 +5610,303 @@ def on_get(self, req, resp):
...
+
+        connection = db.engine.raw_connection()
+        cursor = connection.cursor(db.dict_cursor)
+        cursor.execute('''SELECT `id` FROM `incident` WHERE `active` = 1 AND `bucket_id` IN (SELECT `bucket_id` FROM `IMP_bucket_assignments` WHERE `node_id` = %s)''', node_id)
+        result = cursor.fetchall()
+        cursor.close()
+        connection.close()
+        incident_ids = []
+        for row in result:
+            incident_ids.append(row["id"])
```
jwon, line 5885: you can use a list comprehension here: 
```
incident_ids = [row["id"] for row in result]
```

src/iris/api.py
```diff
@@ -5604,6 +5610,303 @@ def on_get(self, req, resp):
...
+                for bucket in dead_node_buckets:
+                    if i >= len(alive_nodes):
+                        i = 0
+                    cursor.execute("UPDATE IMP_bucket_assignments SET node_id = %s WHERE bucket_id = %s",
+                                   (alive_nodes[i]["node_id"], bucket["bucket_id"]))
+                    i += 1
+
+                #  delete dead nodes from cluster membership
+                cursor.execute(
+                    "DELETE FROM IMP_cluster_members WHERE TIMESTAMPDIFF(SECOND,last_modified, NOW()) >= %s", self.sender_ttl)
```
colinxy, line 5783: Running this delete might get you different result from the previous select because some time has passed. You should just delete based off dead_nodes.

src/iris/utils.py
```diff
@@ -9,17 +9,29 @@
...
+
 uuid4hex = re.compile('[0-9a-f]{32}\Z', re.I)
 allowed_text_response_actions = frozenset(['suppress', 'claim'])
 
 
+def generate_bucket_id():
+    bucket_id_max = config.get("iris-message-processor", {}).get("number_of_buckets", 100)
+    return randrange(bucket_id_max)
```
colinxy, line 32: Document explicitly that bucket_id is a value from 0 to `number_of_bucket-1`, inclusive.

src/iris/utils.py
diegocepedaw: The tests are not run with a config file and this is the only thing in tests that calls load_config() without this load_config() hits a filenotfound error and bails on execution.

src/iris/api.py
```diff
@@ -5604,6 +5610,303 @@ def on_get(self, req, resp):
...
+            # check if node is already a member of the cluster and if new member assign it some buckets
+            cursor.execute("SELECT * FROM IMP_cluster_members where node_id = %s", node_id)
+            if not cursor.fetchone():
+
+                cursor.execute("select * from IMP_cluster_members")
+                cluster_members = cursor.fetchall()
+
+                # node was not previously a member of this cluster so figure out what nodes to assign to itself
+                cursor.execute(
+                    "INSERT INTO IMP_cluster_members VALUES(%s, %s, NOW())", (node_id, "localhost"))
```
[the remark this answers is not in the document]
diegocepedaw: Good catch, that was an oversight that I forgot to change. It should use req.remote_addr for that value.

src/iris/api.py
```diff
@@ -5604,6 +5610,303 @@ def on_get(self, req, resp):
...
+        # configure sql client
+        connection = db.engine.raw_connection()
+        cursor = connection.cursor(db.dict_cursor)
+
+        if self.zk_debug:
+            # use exitstack to replace zookeeper lock context manager for testing
+            lock = ExitStack()
+        else:
+            # configure zookeeper client
+            zk = KazooClient(self.zk_hosts)
```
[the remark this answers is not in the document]
diegocepedaw: It should be possible to do so. I think it shouldn't cause issues with the lock either since it is not re-entrant and repeat calls will block. That being said I was planning on making that change alongside my cache change since they are similar refactoring and this PR is already quite large and complicated.

.ci/setup_mysql.sh
```diff
@@ -11,7 +11,7 @@ mysql -h 127.0.0.1 -u root \
 	-e "SET GLOBAL sql_mode=(SELECT REPLACE(@@sql_mode,'ONLY_FULL_GROUP_BY',''));"
 echo "[*] Setting MySQL max_connections..."
 mysql -h 127.0.0.1 -u root \
-	-e "SET GLOBAL max_connections = 200;"
+	-e "SET GLOBAL max_connections = 500;"
```
diegocepedaw, line 14: consequence of the way the sender cache is initialized by every api process. Will be refactored in next PR so that cache and subsequently the mysql connection they create is shared across all api processes.

src/iris/utils.py
jwon: I'm just confused how this ever worked then. Did something change in this PR that required this?

diegocepedaw: load_config() wasn't being called by the unit tests before until the bucket assignment in utils (and by extension incident creation) used it starting with this PR.

src/iris/api.py
```diff
@@ -5604,6 +5610,301 @@ def on_get(self, req, resp):
...
+
+    def on_get(self, req, resp, node_id):
+
+        connection = db.engine.raw_connection()
+        cursor = connection.cursor(db.dict_cursor)
+        cursor.execute('''SELECT `id` FROM `incident` WHERE `active` = 1 AND `bucket_id` IN (SELECT `bucket_id` FROM `IMP_bucket_assignments` WHERE `node_id` = %s)''', node_id)
+        result = cursor.fetchall()
+        cursor.close()
+        connection.close()
+        incident_ids = []
```
jwon, line 5882: you can delete this line

PR #673, opened 2022-02-24
Create incidents from incoming emails to an OWA mailbox

Fmstrat:
This process monitors an OWA mailbox and creates incidents from those emails.

Also updates `exchangelib` as version 2.2.0 is way out of date and no longer functions.


diegocepedaw:

diegocepedaw (changes requested):

PR #677, opened 2022-03-01
process ldap and oncall syncs concurrently

diegocepedaw:
process ldap and oncall syncs concurrently and add the ability to set the frequency of each sync task independently

### DIFF
--- a/src/iris/bin/sync_targets.py
+++ b/src/iris/bin/sync_targets.py
@@ -47,7 +47,8 @@ stats_reset = {
     'ldap_memberships_removed': 0,
     'ldap_lists_failed_to_add': 0,
     'ldap_memberships_failed_to_add': 0,
-    'ldap_reconnects': 0
+    'ldap_reconnects': 0,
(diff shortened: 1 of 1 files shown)


diegocepedaw:
> We need to seriously consider migrating off of this LDAP library as it doesn't play nice with gevent, so it will definitely block gevent's async loop and may prevent the oncall sync from running concurrently, similar to how it blocks the metrics worker.

@jrgp 100% agree that we need to move off this LDAP library and that is the next step but I think we should separate that change into it's own PR as they are not interdependent and it will be much more involved and it may take me a bit longer to get around to making those changes.  Even with this LDAP library not allowing for proper gevent concurrency because we can set separate run intervals and can run the LDAP sync  task much less often the average time to sync oncall data will be reduced fairly drastically as currently it is blocked by the ldap sync every loop.

jrgp:
@diegocepedaw got it, sounds good

jrgp (approved):

PR #682, opened 2022-03-04
fail build_message reqs when no messages made

diegocepedaw:
- if no messages can be built from notification payload fail with HTTPBadRequest

### DIFF
--- a/src/iris/role_lookup/oncall.py
+++ b/src/iris/role_lookup/oncall.py
@@ -3,21 +3,15 @@
 
 import requests
 import logging
-from iris import metrics
+from iris.metrics import stats
 import oncallclient
 from iris.role_lookup import IrisRoleLookupException
 
 logger = logging.getLogger(__name__)
 
-# metrics
-stats_reset = {
-    'oncall_error': 0
-}
-
 
 class oncall(object):
     def __init__(self, config):
-        metrics.init(config, 'oncall-client', stats_reset)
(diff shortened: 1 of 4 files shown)


src/iris/role_lookup/oncall.py
```diff
@@ -3,15 +3,21 @@
...
 
+# metrics
+stats_reset = {
+    'oncall_error': 0
+}
+
 
 class oncall(object):
     def __init__(self, config):
+        metrics.init(config, 'oncall-client', stats_reset)
```
jrgp, line 20: Do we ever call metrics.emit? 

It looks like the other uses of the metrics functionality in iris are part of the daemons rather than within the webapp uwsgi path.

It may make sense to create a single long lived "metrics" instance and gevent spawn metrics.emit_forever within get_api_app() or get_api()  so only one of them exists per uwsgi worker.

In an ideal scenario though we'd have a metrics aggregator  worker within uwsgi itself that would receive metrics updates from all of the uwsgi workers.

jrgp (changes requested):

src/iris/role_lookup/oncall.py
diegocepedaw: @jrgp that's probably a better long term solution, reverting the metrics changes for now, will implement something like you suggested separately

jrgp (approved):

PR #683, opened 2022-03-07
check that stats are defined before incrementing role_lookup metrics

diegocepedaw:
only sender is correctly configured to emit metrics but api also uses role_lookup. Check if oncall_error is defined in stats before incrementing the metric. Eventually we should set up a mechanism for api to allow emitting metrics at is it currently not configured to directly any.

### DIFF
--- a/src/iris/role_lookup/oncall.py
+++ b/src/iris/role_lookup/oncall.py
@@ -3,21 +3,15 @@
 
 import requests
 import logging
-from iris import metrics
+from iris.metrics import stats
 import oncallclient
 from iris.role_lookup import IrisRoleLookupException
 
 logger = logging.getLogger(__name__)
 
-# metrics
-stats_reset = {
-    'oncall_error': 0
-}
-
 
 class oncall(object):
     def __init__(self, config):
-        metrics.init(config, 'oncall-client', stats_reset)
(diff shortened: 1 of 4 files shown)


src/iris/role_lookup/oncall.py
```diff
@@ -28,7 +28,8 @@ def call_oncall(self, url, expected_type):
         try:
             r = self.requests.get(url)
         except Exception:
-            stats['oncall_error'] += 1
```
jrgp, line 31: Could we avoid these checks by making `stats` a colletions.defaultdict(int) or a collections.Counter()?

jrgp (changes requested):

src/iris/role_lookup/oncall.py
diegocepedaw: @jrgp we could, but is there an advantage? As it stands right now when api.py uses oncall.py through role_lookup the stats dictionary will be empty and never used for anything since api.py is not set up to emit metrics (unlike sender.py). Is there any use to make it so we can increment a field in a dictionary that is never going to be used? My thought was that this way it is clear that this is a metric that is not always used at the moment. That being said I wouldn't particularly mind doing it either way.

jrgp (approved):

PR #693, opened 2022-03-29
use locks with api cache

diegocepedaw:
add a threading lock to the api cache to fix a bug where request from known good application can fail authentication with `Tried authenticating with nonexistent app: "app_foo"` when api cache is being updated.

### DIFF
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -701,7 +701,10 @@ def is_valid_tracking_settings(t, k, tpl):
                 except jinja2.TemplateSyntaxError as e:
                     return False, 'Invalid jinja syntax in email html: %s' % e
     else:
-        if t not in cache.modes:
+        cache.api_cache_lock.acquire()
+        mode = cache.modes.get(t)
+        cache.api_cache_lock.release()
+        if not mode:
             return False, 'Unknown tracking type: %s' % t
 
         environment = SandboxedEnvironment()
@@ -793,7 +796,9 @@ class AuthMiddleware(object):
         # API call, but set 'app' to the internal iris user as some routes (test incident creation)
         # need it.
         if req.context['username']:
+            cache.api_cache_lock.acquire()
             req.context['app'] = cache.applications.get('iris')
+            cache.api_cache_lock.release()
             return
 
         # For the purpose of e2etests, allow setting username via header, rather than going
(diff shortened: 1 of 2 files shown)


src/iris/cache.py
```diff
@@ -3,6 +3,7 @@
 
 from . import db
 import logging
+import threading
```
jrgp, line 6: We should use gevent.Lock instead of threading.Lock

src/iris/api.py
```diff
@@ -701,7 +701,10 @@ def is_valid_tracking_settings(t, k, tpl):
                 except jinja2.TemplateSyntaxError as e:
                     return False, 'Invalid jinja syntax in email html: %s' % e
     else:
-        if t not in cache.modes:
+        cache.api_cache_lock.acquire()
+        mode = cache.modes.get(t)
+        cache.api_cache_lock.release()
```
jrgp, line 706: The Lock object should support acting like a context manager. We can replace the manual acquire/release calls with:

```python
with cache.api_cache_lock:
    mode = cache.modes.get(t)
```

jrgp (changes requested):

src/iris/cache.py
jwon: does it matter though since we're monkey patching?

jrgp: Sorry, I made a typo: we should use `gevent.lock.Semaphore`. 

It looks like gevent does monkey patch the threading module per `https://www.gevent.org/api/gevent.monkey.html#gevent.monkey.patch_thread` but when we interact with these primitives directly it may be more clear/explicit to use the gevent equivalents, and not rely on the monkey patch ordering to fix this for us (as afaik in our internal uwsgi wrapper we only patch ssl to start with)

This would also be consistent with how we're doing locking in the rate limiting module: https://github.com/linkedin/iris/blob/master/src/iris/sender/quota.py

diegocepedaw: like James said it gets monkey patched but it does seem clearer and more consistent to use the gevent lock explicitly. Plus we can avoid any weird bugs in the future.

src/iris/api.py
diegocepedaw: cool, I hadn't realized the gevent semaphore is a context manager and can be used in with statements

jrgp (approved):

src/iris/cache.py
```diff
@@ -46,7 +48,8 @@ def cache_applications():
                           WHERE `application_id` = %s''', app['id'])
         app['categories'] = {row['name']: row for row in cursor}
         new_applications[app['name']] = app
-    applications = new_applications
+    with api_cache_lock:
+        applications = new_applications
```
jwon, line 52: huh? do we need a lock here?

PR #700, opened 2022-04-06
add render jinja endpoint

diegocepedaw:
create internal endpoint to flexibly render a raw jinja2 template string with an arbitrary context

### DIFF
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -1607,6 +1607,11 @@ class Incidents(object):
                             incident_IDs.append(message.get('incident_id'))
                         where.append('''`incident`.`id` IN %s''')
                         sql_values.append(tuple(incident_IDs))
+                    elif target:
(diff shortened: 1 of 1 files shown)


src/iris/api.py
```diff
@@ -5847,6 +5847,36 @@ def on_delete(self, req, resp, username, application):
         resp.status = HTTP_204
 
 
+class InternalRenderJinja():
```
jrgp, line 5850: Will this also render the markdown syntax used in email templates?

jrgp (changes requested):

src/iris/api.py
diegocepedaw: @jrgp no, the markdown will be rendered separately within iris-message-processor. This endpoint is exclusively to render tracking notification jinja (I still have to make an RB for that change).

jwon (approved):

Issue #699, opened 2022-04-05
Create Plan Fail


bla-ckbox:
I still encounter issues below when testing with the **master** branch commit https://github.com/linkedin/iris/commit/500822ac5d757884246c95145d6a89531c503f81  

- `Error: Invalid plan - Priority not found for step 1` when saving plan
- Error 400 `Target type user not found` on /v0/targets/user

_sync_targets.py_ seems to be running fine:
```
2022-04-05 16:12:44,234 INFO root Sending metrics: failed_tasks: 0, ldap_lists_added: 0, ldap_lists_failed_to_add: 0, ldap_lists_found: 0, ldap_lists_removed: 0, ldap_memberships_added: 0, ldap_memberships_failed_to_add: 0, ldap_memberships_found: 0, ldap_memberships_removed: 0, ldap_reconnects: 0, others_purged: 0, sql_errors: 0, teams_added: 0, teams_failed_to_add: 0, teams_found: 0, user_contacts_updated: 0, users_added: 0, users_failed_to_add: 0, users_failed_to_update: 0, users_found: 0, users_purged: 0
2022-04-05 16:12:44,234 INFO root Starting oncall sync...
2022-04-05 16:12:44,441 INFO root Users to insert (0)
2022-04-05 16:12:44,441 INFO root Users to update (5)
2022-04-05 16:12:44,447 INFO root Teams to insert (0)
2022-04-05 16:12:44,450 INFO root Users to mark inactive (0)
2022-04-05 16:12:44,450 INFO root Teams to mark inactive (0)
2022-04-05 16:12:44,451 INFO root oncall sync took 0.22 seconds
```

_target_type_, _target_, and _priority_ are not empty
```
mysql> select * from target_type;
+----+--------------+
| id | name         |
+----+--------------+
|  3 | mailing-list |
|  2 | team         |
|  1 | user         |
+----+--------------+
3 rows in set (0.00 sec)

mysql> select * from target;
+----+--------------------------------------+---------+--------+
| id | name                                 | type_id | active |
+----+--------------------------------------+---------+--------+
|  1 | demo                                 |       1 |      0 |
|  4 | b291694f-7789-4436-a5d2-ca37f67b3df3 |       2 |      0 |
|  5 | 8558d07b-b5ae-44c3-b3c0-6f4179e79402 |       2 |      0 |
|  6 | abc                                  |       3 |      1 |
|  7 | demo                                 |       3 |      1 |
|  8 | jp******                             |       1 |      1 |
[...]

Issue #644, opened 2021-10-19
iris Plan issue 


allwyn-pradip:
with the new images, new or existing plans do not get saved. 

the error pops up as 

Error: Invalid plan - Priority not found for step 1

even when the priority are setup

![image](https://user-images.githubusercontent.com/56670753/137998096-37ba4e28-fd08-4976-ad58-0edabfd9bead.png)

allwyn-pradip:
@diegocepedaw any update what is causing this issue.

diegocepedaw:
@allwyn-pradip unfortunately I have not had much bandwidth recently to look at community issues lately. Do you remember which release version this issue was introduced in?

allwyn-pradip:
@diegocepedaw I am at the latest version 1.36 testing it for python3

diegocepedaw:
@allwyn-pradip what do you get if you run `SELECT `name`, `id` FROM target_type` in your mysql? That error you linked means Iris didn't find the type user in its cache, and the cache should be pulling the types from the database.

mohammad-sajid:
@diegocepedaw  the output for the query is fine. 
MySQL [iris]> SELECT name, id FROM target_type;
+--------------+----+
| name         | id |
+--------------+----+
| mailing-list |  3 |
| team         |  2 |
| user         |  1 |
+--------------+----+

Also facing the issue Error: Invalid plan - Target NOC-Team not found for step 1. 
But we have this team created at oncall DB. 
what will be the reason, why its not taking the team names from oncall DB. ?

![image](https://user-images.githubusercontent.com/57441817/152889799-cb3e797d-39ce-42a7-9d8a-28d0136ad56c.png)

diegocepedaw:
Is the sync targets script running and correctly able to git oncall-api?

mohammad-sajid:
@diegocepedaw  what is the sync target script? 
where can I find that oncall-api ? 
could you please help here.

mohammad-sajid:
@diegocepedaw  i saw there is one sync_target.py code is available in srs/iris/bin. executed manually and getting below error. could you please help here. 

2022-02-08 11:26:10,489 ERROR root Error syncing from oncall!
Traceback (most recent call last):
  File "sync_targets.py", line 786, in main
    sync_from_oncall(config, engine)
  File "sync_targets.py", line 174, in sync_from_oncall
    oncall_users = fetch_users_from_oncall(oncall)
  File "sync_targets.py", line 157, in fetch_users_from_oncall
[...]

PR #703, opened 2022-04-11
Call gevent.monkey.patch_all immediately on import

### LINKED ISSUE #686: Unable to publish new (or existing) Plan
I'm currently trying to setup iris, but I'm ending up with an error while adding a new plan or saving an existing plan after installatio. I tried with latest master as well as with latest released version.

Screenshot 2022-03-18 at 14 47 37

Easiest steps to reproduce:
- Use the code from #684 to run iris inside a docker compose setup.
- Update `config.dev.yaml` to use the correct database
- Go to localhost:16649, login with demo/demo
- Try to add a new plan (see screenshot).

bilbof:
The api makes use of [gevent], a coroutine based networking library which relies heavily on monkey patching the stdlib.

From the [gevent.monkey] docs:
> Warning Patching too late can lead to unreliable behaviour
> (for example, some modules may still use blocking sockets) or even errors.

This appears to have happened here. Thanks to @allwyn-pradip for pointing me at the right file in PR https://github.com/linkedin/iris/pull/690!

Monkey-patching immediately resolves #686, #699, #644.

Blog on gevent: https://eng.lyft.com/what-the-heck-is-gevent-4e87db98a8
> In the case of gevent — monkey patching has to be the absolute first thing a process does

[gevent]: https://www.gevent.org/index.html
[gevent.monkey]: https://www.gevent.org/api/gevent.monkey.html

### DIFF
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -1,6 +1,7 @@
 # Copyright (c) LinkedIn Corporation. All rights reserved. Licensed under the BSD-2 Clause license.
 # See LICENSE in the project root for license information.
-from gevent import spawn, sleep, socket, Timeout
+from gevent import spawn, sleep, socket, Timeout, monkey
+monkey.patch_all() # NOQA
 
 import msgpack
 import time


allwyn-pradip:
@diegocepedaw -- yes it is working as expected.

PR #724, opened 2022-06-20
fix: only allow proper prometheus metric names

roock:
Metric Names are documented here:
https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels

### DIFF
--- a/src/iris/metrics/prometheus.py
+++ b/src/iris/metrics/prometheus.py
@@ -33,6 +33,13 @@ class prometheus(object):
             return
         for metric, value in metrics.items():
             if metric not in self.gauges:
-                self.gauges[metric] = Gauge(self.appname + '_' + metric, '')
+                metricname = metric
+                # create a valid prometheus metric name
+                # replace spaces and hyphens with underscores to make the metric more readable
+                # afterwards remove any other character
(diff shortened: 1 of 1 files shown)


src/iris/metrics/prometheus.py
```diff
@@ -33,6 +33,13 @@ def send_metrics(self, metrics):
             return
         for metric, value in metrics.items():
             if metric not in self.gauges:
-                self.gauges[metric] = Gauge(self.appname + '_' + metric, '')
+                metricname = metric
+                # create a valid prometheus metric name
+                # replace spaces and hyphens with underscores to make the metric more readable
+                # afterwards remove any other character
+                metricname = re.sub('-', "_", metricname)
+                metricname = re.sub(' ', "_", metricname)
```
jrgp, line 41: When we're doing straightforward string replacements that aren't using regexes, we should prefer the standard `str.replace()` method, as this is faster and we avoid the risk of working against special regex symbols.

jrgp (changes requested):

src/iris/metrics/prometheus.py
roock: @jrgp Great point! Fixed!

PR #726, opened 2022-07-14
send messages even if they fail to render

diegocepedaw:
- fix bugs where messages that fail to render would return a 400 to the external sender instead of the unrendered message
- fix bug where message reprioritization would cause message to fail to render

### DIFF
--- a/src/iris/bin/sender.py
+++ b/src/iris/bin/sender.py
@@ -1010,7 +1010,7 @@ def render(message):
             # format define a generic template that will work for all contexts
             # - even those that are invalid as a final final format maybe repr or pprint
             logger.error(error, message)
-            message['subject'] = 'Iris failed to render your message' % message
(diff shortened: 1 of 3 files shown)


src/iris/bin/sender.py
```diff
@@ -1010,13 +1010,13 @@ def render(message):
             # format define a generic template that will work for all contexts
             # - even those that are invalid as a final final format maybe repr or pprint
             logger.error(error, message)
-            message['subject'] = '%(message_id)s Iris failed to render your message' % message
+            message['subject'] = 'Iris failed to render your message' % message
```
jrgp, line 1013: Should we remove the `% message` because we're no longer interpolating any params into the string?

diegocepedaw: don't know how I missed that, done

jrgp (approved):

PR #728, opened 2022-08-31
fix message destination deduplication

diegocepedaw:
previous fix did not account for mode so would deduplicate sms and calls to the same destination

### DIFF
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -6081,19 +6081,13 @@ class InternalBuildMessages():
 
         # deduplicate messages with same destination
         unique_messages = []
-        unique_destinations = {}
+        unique_destinations = defaultdict(set)
         for message in messages:
-            if message.get("destination") and message.get("mode") not in unique_destinations.get(message["destination"], []):
-                if message["destination"] not in unique_destinations:
-                    unique_destinations[message["destination"]] = [message["mode"]]
-                else:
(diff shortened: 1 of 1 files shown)


src/iris/api.py
```diff
@@ -6081,13 +6081,19 @@ def on_get(self, req, resp):
...
-        unique_destinations = set()
+        unique_destinations = {}
         for message in messages:
-            if message.get("destination") and message.get("destination") not in unique_destinations:
-                unique_destinations.add(message["destination"])
+            if message.get("destination") and message.get("mode") not in unique_destinations.get(message["destination"], []):
+                if message["destination"] not in unique_destinations:
+                    unique_destinations[message["destination"]] = [message["mode"]]
+                else:
+                    unique_destinations[message["destination"]] = unique_destinations[message["destination"]].append(message["mode"])
```
jrgp, line 6090: We could avoid a bit of boilerplate by using a defaultdict(set) instead of a dict.

Eg

```python
unique_destinations = defaultdict(set)
for message in messages:
  if message.get("destination") and message.get("mode") not in unique_destinations[message["destination"]]:
     unique_destinations[message["destination"]].add(message["mode"])
# and so on
```

jrgp (changes requested):

jrgp (approved):

PR #729, opened 2022-10-27
allow use of custom incident creation handler for webhook endpoints

diegocepedaw:


src/iris/webhooks/grafana.py
```diff
@@ -1,19 +1,25 @@
...
 
 logger = logging.getLogger(__name__)
 
 
-class grafana(object):
+class grafana(webhook):
     allow_read_no_auth = False
 
+    def __init__(self, config):
+        self.custom_incident_handler_dispatcher = CustomIncidentHandlerDispatcher(config)
```
jrgp, line 19: Do we need to define this constructor? It seems to be a duplicate of the inherited constructor from `webhook`, and the other web hook implementations that inherit from `webhook` are still constructor-less.

src/iris/webhooks/grafana.py
```diff
@@ -87,3 +93,17 @@ def on_post(self, req, resp):
...
+        if self.custom_incident_handler_dispatcher.handlers:
+            incident_data = {
+                'id': incident_id,
+                'plan': plan,
+                'plan_id': plan_id,
+                'created': int(time.time()),
+                'application': app,
+                'context': alert_params
+            }
+            super().custom_incident_handler(incident_data)
```
jrgp, line 109: Instead of `super()` could we use `self`?

src/iris/webhooks/webhook.py
```diff
@@ -3,17 +3,23 @@
 import datetime
 import logging
 import ujson
+import time
 from falcon import HTTP_201, HTTPBadRequest, HTTPNotFound
 
 from iris import db
 from iris import utils
+from iris.custom_incident_handler import CustomIncidentHandlerDispatcher
+from iris.constants import (PRIORITY_PRECEDENCE_MAP)
```
jrgp, line 12: Can we remove the parens surrounding PRIORITY_PRECEDENCE_MAP?

jrgp (changes requested):

jrgp (approved):

PR #739, opened 2023-03-02
Adding additional context for triaging

cshelke52889:
Added list of plan ids which we can add in the allow_list and a config enable_triage_context for enabling everywhere.
This will be disabled when we allow only from the allow_list.
Later on we can just enable enable_triage_context for everywhere.

Testing
- make unit
- make serve
- Tested in local on local iris db and verified the new context

### DIFF
--- a/configs/config.dev.yaml
+++ b/configs/config.dev.yaml
@@ -194,6 +194,12 @@ oncall-key: abc # API key for oncall auth
 applications:
  - dummy_app
 
+## Enable addition of metavariables to the incident context by default.
+## If disabled, the metavariables will only added in the incident context for plans in the triage_allow_list.
+enable_default_metavariables_in_context: False
+triage_allow_list:
+- dummy_item_1
+
 ## Vendors provides a way to add support for new notification mechanisms
 vendors: []
 #- type: iris_smtp
--- a/configs/config.dev.yaml
+++ b/configs/config.dev.yaml
@@ -195,9 +195,9 @@ applications:
  - dummy_app
 
 ## Enable addition of metavariables to the incident context by default.
-## If disabled, the metavariables will only added in the incident context for plans in the triage_allow_list.
+## If disabled, the metavariables will only be added in the incident context for plans in the allow list.
 enable_default_metavariables_in_context: False
-triage_allow_list:
+metavariables_in_context_allow_list:
 - dummy_item_1
 
 ## Vendors provides a way to add support for new notification mechanisms
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -1512,7 +1512,7 @@ class Incidents(object):
         self.custom_incident_handler_dispatcher = CustomIncidentHandlerDispatcher(config)
         # if True, we enable metavariables in the context everywhere, if False they will be enabled only for plans in allow list
         self.enable_default_metavariables_in_context = config.get('enable_default_metavariables_in_context', False)
-        self.triage_allow_list = config.get('triage_allow_list', [])
+        self.metavariables_in_context_allow_list = config.get('metavariables_in_context_allow_list', [])
 
     def on_get(self, req, resp):
         '''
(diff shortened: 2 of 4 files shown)


src/iris/api.py
```diff
@@ -1791,6 +1791,13 @@ def on_post(self, req, resp):
                                                    `current_step`, `active`, `application_id`, `bucket_id`)
                            VALUES (:plan_id, :created, :context, 0, :active, :application_id, :bucket_id)''',
                         data).lastrowid
+                    # adding additional context
+                    if self.enable_triage_context or plan_id in self.triage_allow_list:
+                        context['incident_id'] = incident_id
```
diegocepedaw, line 1796: these iris context variables are not meant to exist in the top level of the context but rather inside the 'iris' key within the context i.e. context['iris']['incident_id']

src/iris/api.py
```diff
@@ -1791,6 +1791,13 @@ def on_post(self, req, resp):
...
+                        context['created'] = data['created']
```
diegocepedaw, line 1797: if we're going to add these iris variables here we should add all the built-in iris context variables as well

iris.target (target user name)
iris.priority
iris.application (application name)
iris.plan (plan name)
iris.plan_id

iris.incident_id

iris.template (template name)

diegocepedaw (changes requested):

src/iris/api.py
```diff
@@ -1791,6 +1791,13 @@ def on_post(self, req, resp):
                                                    `current_step`, `active`, `application_id`, `bucket_id`)
                            VALUES (:plan_id, :created, :context, 0, :active, :application_id, :bucket_id)''',
                         data).lastrowid
```
diegocepedaw, line 1793: fix the flake8 errors and make sure the circleci tests are passing

diegocepedaw (changes requested):

configs/config.dev.yaml
```diff
@@ -194,6 +194,10 @@ oncall-key: abc # API key for oncall auth
 applications:
  - dummy_app
 
+enable_triage_context: False
```
diegocepedaw, line 197: rename config to something more descriptive of what it does (like enable_default_metavariables_in_context) and add a comment explaining config option

diegocepedaw (changes requested):

src/iris/api.py
```diff
@@ -1791,6 +1791,13 @@ def on_post(self, req, resp):
...
+                    # adding additional context
+                    if self.enable_triage_context or plan_id in self.triage_allow_list:
+                        context['incident_id'] = incident_id
+                        context['created'] = data['created']
+                        context_json_str = ujson.dumps({variable: context.get(variable)
+                                            for variable in app['variables']})
+                        session.execute('''UPDATE `incident` SET `context` = :context_json_str where `id` = :incident_id ''',{'context_json_str': context_json_str, 'incident_id': incident_id})
```
diegocepedaw, line 1800: This extra context should also be added to all endpoints that create incidents like hooks and create incidents from email

diegocepedaw (changes requested):
T

src/iris/api.py
cshelke52889: added

src/iris/api.py
cshelke52889: added

configs/config.dev.yaml
cshelke52889: done

src/iris/api.py
cshelke52889: fixed

configs/config.dev.yaml
```diff
@@ -194,6 +194,12 @@ oncall-key: abc # API key for oncall auth
 applications:
  - dummy_app
 
+## Enable addition of metavariables to the incident context by default.
+## If disabled, the metavariables will only added in the incident context for plans in the triage_allow_list.
+enable_default_metavariables_in_context: False
+triage_allow_list:
```
diegocepedaw, line 200: nit: since we renamed the config lets also rename the allow list. Otherwise everything els lgtm

diegocepedaw, line 200: nit: since we renamed the config lets also rename the allow list maybe (metavariables_in_context_allow_list) or something. Otherwise everything else lgtm

diegocepedaw (changes requested):

PR #742, opened 2023-05-30
enable zookeeper to use ssl for external sender heartbeat

diegocepedaw:
add optional config to use ssl for connecting to zookeeper for external_sender heartbeat

### DIFF
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -6227,11 +6227,6 @@ class SenderHeartbeat():
         self.number_of_buckets = cfg.get("number_of_buckets", 100)
         self.sender_ttl = cfg.get("sender_ttl", 60)
         self.zk_debug = cfg.get("zk_debug", True)
-        self.zk_cluster=cfg.get("zookeeper_cluster"),
-        self.zk_use_ssl=cfg.get('zk_use_ssl'),
-        self.certfile=cfg.get('ssl_certificate'),
-        self.keyfile=cfg.get('ssl_certificate_key'),
-        self.ca_path=cfg.get('ca_bundle_path')
         self.zk_client = KazooClient(
             hosts=cfg.get("zookeeper_cluster"),
             use_ssl=cfg.get('zk_use_ssl', False),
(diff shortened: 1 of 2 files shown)


src/iris/api.py
```diff
@@ -6219,12 +6221,22 @@ class SenderHeartbeat():
...
-        cfg = config.get("iris-message-processor", {})
-        self.zk_hosts = cfg.get("zk_hosts", "127.0.0.1:2181")
+        cfg = config.get("external_sender", {})
         self.lock_path = cfg.get("lock_path", "/iris_message_processor/db_locks/bucket_lock")
         self.number_of_buckets = cfg.get("number_of_buckets", 100)
-        self.sender_ttl = config.get("sender_ttl", 60)
-        self.zk_debug = config.get("zk_debug", True)
+        self.sender_ttl = cfg.get("sender_ttl", 60)
+        self.zk_debug = cfg.get("zk_debug", True)
+        if not self.zk_debug:
```
colinxy, line 6229: This means when zk_debug=True, we won't start kazoo client. Is it expected?

src/iris/api.py
```diff
@@ -6219,12 +6221,22 @@ class SenderHeartbeat():
...
 
     def __init__(self, config):
-        cfg = config.get("iris-message-processor", {})
-        self.zk_hosts = cfg.get("zk_hosts", "127.0.0.1:2181")
+        cfg = config.get("external_sender", {})
         self.lock_path = cfg.get("lock_path", "/iris_message_processor/db_locks/bucket_lock")
         self.number_of_buckets = cfg.get("number_of_buckets", 100)
-        self.sender_ttl = config.get("sender_ttl", 60)
-        self.zk_debug = config.get("zk_debug", True)
+        self.sender_ttl = cfg.get("sender_ttl", 60)
```
colinxy, line 6227: We moved sender_ttl and zk_debug from top level keys to keys under external_sender, which is a breaking change. Is that expected?

src/iris/api.py
diegocepedaw: yeah, this option means within SenderHeartbeat() exitstack will to replace zookeeper lock context manager for testing so there is no need for a kazoo client to be started if this option is enabled

PR #754, opened 2023-10-17
add name_only param to plan,template,application search 

diegocepedaw:
facilitate typeahead searches by adding `name_only` param that returns just a flat list of names instead of all the data about plans, templates, and applications


src/iris/api.py
```diff
@@ -2809,6 +2819,16 @@ def on_get(self, req, resp):
...
 
+        if req.get_param_as_bool('name_only', default=False):
+            flat_name_list = []
+            for template in cursor:
+                flat_name_list.append(template.get('name'))
+            connection.close()
+            payload = ujson.dumps(flat_name_list)
+            resp.status = HTTP_200
+            resp.body = payload
+            return
```
tbrockman, line 2830: Is this equivalent to adding `fields` and only specifying `name`?

PR #755, opened 2023-10-17
make application api filterable

diegocepedaw:
add the ability to filter application and return only specific fields

### DIFF
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -572,6 +572,40 @@ application_filter_types = {
 
 application_query = '''SELECT %s FROM `application`'''
 
+get_all_vars_query = 'SELECT `application`.`id` as app_id, `template_variable`.`name` as name, `template_variable`.`required` as required, `template_variable`.`title_variable` as title_variable FROM `template_variable` JOIN `application` on `template_variable`.`application_id` = `application`.`id`'
+
+get_all_default_application_modes_query = '''
+SELECT `application`.`id` as app_id, `priority`.`name` as priority, `mode`.`name` as mode
+FROM `default_application_mode`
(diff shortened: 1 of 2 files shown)


src/iris/api.py
```diff
@@ -4036,36 +4056,57 @@ class Applications(object):
...
+                app['supported_modes'] = [row['name'] for row in cursor]
+
+                cursor.execute(get_application_owners_query, app['id'])
+                app['owners'] = [row['name'] for row in cursor]
+
+                cursor.execute(get_application_custom_sender_addresses, app['id'])
+                app['custom_sender_addresses'] = {row['mode_name']: row['address'] for row in cursor}
+
+                cursor.execute(get_application_categories, app['id'])
+                app['categories'] = [row for row in cursor]
```
jrgp, line 4107: This looks like it will run these 5 queries for every application returned by the above the query serially, if requested_fields is not specified as a url param? 

As we have lots of applications configured, and as afaik this endpoint is hit frequently to populate things like app names in typeaheads/etc, I'm concerned this will make things noticeably slow.

Is it possible to instead do something like:

- After we run `apps = cursor.fetchall()`, go through and get every ID returned 
- afterwards, run all of these get_application_*_query queries but using "where app id in LISTOFIDS" 
- go through those result sets and interpolate them into the apps list we got back on line 4079?

This way we will only be running 6 queries total instead of 1 + (5*268)

diegocepedaw: This was already being run every time for this endpoint, I'm guessing since originally we had very few apps whoever wrote it was not very concerned about performance. I was originally planning to optimize it when I made these additional fields selectable once I would be changing this code then. I had some time so I went ahead and implemented that optimization as well as the TODO of making these other fields individually selectable

jrgp (approved):

PR #759, opened 2023-12-07
add counts to search api

diegocepedaw:
Add the ability to specify the parameter `counts` in incident, plan, and template search to efficiently return a dictionary with the unique field values and the number of rows that have those values. Incident example below:

```
**Example request**:
.. sourcecode:: http
    GET /v0/incidents?created__ge=1700693640&fields=owner&fields=application&counts=true  HTTP/1.1
    Host: example.com
**Example response**:
.. sourcecode:: http
    HTTP/1.1 200 OK
    Content-Type: application/json
    {
        "field_counts": {
            "owner": {
                "foo_user": 4,
                "None": 7
            },
            "application": {
                "Autoalerts": 11
            }
        },
        "total_count": 11
    }
```

### DIFF
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -11,7 +11,7 @@ import random
 import re
 import time
 import uuid
-from collections import defaultdict
+from collections import Counter, defaultdict
 from contextlib import ExitStack
 from urllib.parse import parse_qs
 
@@ -831,12 +831,12 @@ def gen_where_filter_clause(connection, filters, filter_types, kwargs):
     return where
 
 
-def generate_grouped_query(query, columns, fields):
+def generate_grouped_query(query, columns, allowed_fields):
     '''modify query to retrieve total counts for distinct combinations of specified columns while avoiding the return of each unique row individually'''
     # avoid DISTINCT grouping
     modified_query = query.replace("SELECT DISTINCT", "SELECT")
     # Construct the counts query
-    group_by_clause = ", ".join([f"`{column}`" for column in fields if column in columns])
+    group_by_clause = ", ".join([f"`{column}`" for column in allowed_fields if column in columns])
     count_query = f"SELECT {group_by_clause}, COUNT(*) as `group_count` FROM ({modified_query}) AS subquery GROUP BY {group_by_clause}"
     return count_query
 
@@ -844,7 +844,7 @@ def generate_grouped_query(query, columns, fields):
 def format_count_results(results):
     '''parse mysql results into counts dictionary'''
     id_count = 0
-    count_dict = defaultdict(lambda: defaultdict(int))
+    count_dict = defaultdict(Counter)
     for idx, item in enumerate(results):
         for key, value in item.items():
             # group count contains only the counts, handle differently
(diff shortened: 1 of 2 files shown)


src/iris/api.py
```diff
@@ -831,6 +831,41 @@ def gen_where_filter_clause(connection, filters, filter_types, kwargs):
     return where
 
 
+def generate_grouped_query(query, columns, fields):
+    '''modify query to retrieve total counts for distinct combinations of specified columns while avoiding the return of each unique row individually'''
+    # avoid DISTINCT grouping
+    modified_query = query.replace("SELECT DISTINCT", "SELECT")
+    # Construct the counts query
+    group_by_clause = ", ".join([f"`{column}`" for column in fields if column in columns])
```
jrgp, line 839: Can we add validation that at least one valid field was provided and 4xx otherwise, to avoid a broken query?

src/iris/api.py
```diff
@@ -1267,11 +1302,54 @@ def on_get(self, req, resp):
...
+                        "user_foo": 10,
+                        "user_bar": 20,
+                        "user_baz": 5
+                    }
+                },
+                "total_count": 35
+            }
+
         '''
+        counts_only = req.get_param('counts')
```
jrgp, line 1351: Could this be get_param_as_bool() instead?

src/iris/api.py
```diff
@@ -831,6 +831,41 @@ def gen_where_filter_clause(connection, filters, filter_types, kwargs):
     return where
 
 
+def generate_grouped_query(query, columns, fields):
```
jrgp, line 834: Columns and fields are kind of synonyms which make the readability here difficult. 

Could we rename one of them to "allowed_fields" or similar to more easily reflect their usage as a field allowlist?

src/iris/api.py
```diff
@@ -831,6 +831,41 @@ def gen_where_filter_clause(connection, filters, filter_types, kwargs):
...
+    # Construct the counts query
+    group_by_clause = ", ".join([f"`{column}`" for column in fields if column in columns])
+    count_query = f"SELECT {group_by_clause}, COUNT(*) as `group_count` FROM ({modified_query}) AS subquery GROUP BY {group_by_clause}"
+    return count_query
+
+
+def format_count_results(results):
+    '''parse mysql results into counts dictionary'''
+    id_count = 0
+    count_dict = defaultdict(lambda: defaultdict(int))
```
jrgp, line 847: Could we replace the inner lambda with a collections.Counter? 

Eg defaultdict(Counter) to get the same effect.

src/iris/api.py
diegocepedaw: added

src/iris/api.py

src/iris/api.py
diegocepedaw: changed to make it clearer

src/iris/api.py

src/iris/api.py
```diff
@@ -1665,7 +1764,41 @@ def on_get(self, req, resp):
...
+                        "None": 7
+                    },
+                    "application": {
+                        "Autoalerts": 11
+                    }
+                },
+                "total_count": 11
+            }
         '''
+        counts_only = req.get_param('counts')
```
jrgp, line 1800: Would it be possible for the other get_param calls added in this PR also get converted to get_param_as_bool?

diegocepedaw: ah, yeah

jrgp (approved):

PR #761, opened 2024-01-16
optional incident rate limiting

diegocepedaw:
Add the ability to optionally enforce rate limits for creating incidents per application per plan

max_active_incidents: sets the maximum allowable active incidents for a given app+plan
max_recent_incidents: how many incidents maximum are allowed to be created within a time window
recent_lookback_seconds: time window in seconds for max_recent_incidents lookup
exempt_applications: allows specifying applications that can ignore the rate-limit and create unlimited incidents

### DIFF
--- a/configs/config.dev.yaml
+++ b/configs/config.dev.yaml
@@ -116,6 +116,15 @@ webhooks:
 #   external_sender_version: 0
 #   zookeeper_cluster: localhost:2181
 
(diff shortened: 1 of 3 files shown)


jrgp (approved):
LGTM. 

Could we also add tests at some point?

tbrockman (approved):

PR #762, opened 2024-01-18
Avoid 5xx's when creating templates/plans with invalid targets

jrgp:
Catch these integrity errors and return them as 400s instead of raising 500s.

Needed when importing *lots* of templates/plans in bulk where a majority of them will reference invalid targets.


diegocepedaw (approved):
LGTM

PR #764, opened 2024-02-12
Add support for Oncall 'director' role lookup

diegocepedaw:
Add support for Oncall 'director' role lookup

### DIFF
--- a/src/iris/role_lookup/oncall.py
+++ b/src/iris/role_lookup/oncall.py
@@ -80,6 +80,12 @@ class oncall(object):
                 return [user['user'] for user in result]
             else:
                 return None
+        elif role == 'director':
+            result = self.call_oncall('/teams/%s/oncall/director' % target, list)
+            if result:
+                return [user['user'] for user in result]
+            else:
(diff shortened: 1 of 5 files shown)


src/iris/role_lookup/oncall.py
```diff
@@ -80,6 +80,12 @@ def get(self, role, target):
                 return [user['user'] for user in result]
             else:
                 return None
+        elif role == 'director':
+            result = self.call_oncall('/teams/%s/oncall/director' % target, list)
+            if result:
+                return [user['user'] for user in result]
+            else:
+                return None
```
tbrockman:
```suggestion
        elif role in ['director', 'manager']:
            result = self.call_oncall('/teams/%s/oncall/%s' % (target, role), list)
            if result:
                return [user['user'] for user in result]
            else:
                return None
```

Or something like it might bit a bit cleaner.

diegocepedaw: changed

tbrockman (approved):

PR #765, opened 2024-04-23
limit filter query 'in' length

diegocepedaw:
return a 400 error if a client provides more than 100 items in a filter query 'in' search to guard against too large SQL queries which can cause 500s

### DIFF
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -11,7 +11,7 @@ import random
 import re
 import time
 import uuid
-from collections import defaultdict
+from collections import Counter, defaultdict
 from contextlib import ExitStack
 from urllib.parse import parse_qs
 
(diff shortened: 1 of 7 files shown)


src/iris/api.py
```diff
@@ -1350,6 +1366,7 @@ def on_get(self, req, resp):
             }
 
         '''
+        check_param_list_len(req.params)
```
jrgp: Could this be called from one of the middleware functions common to all routes, rather than copy pasted into each route handler?

diegocepedaw: yeah, I think it is used widely enough, and has minimal enough impact for endpoints it doesn't apply to that it makes sense, changed

jrgp (approved):

PR #766, opened 2024-06-05
Add application template search

diegocepedaw:
Add the ability to search for all the active templates that have a definition for a given application.

### DIFF
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -11,7 +11,7 @@ import random
 import re
 import time
 import uuid
-from collections import defaultdict
+from collections import Counter, defaultdict
 from contextlib import ExitStack
 from urllib.parse import parse_qs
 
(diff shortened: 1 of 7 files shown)


src/iris/api.py
```diff
@@ -4472,6 +4472,79 @@ def on_get(self, req, resp, app_name):
...
+                   JOIN `application` ON `template_content`.`application_id` = `application`.`id`
+                   JOIN `target` ON `target`.`id` = `template`.`user_id`
+                   WHERE %s
+                   GROUP BY `template`.`id`""" % (
+            ",".join(template_columns[f] for f in fields if f in template_columns),
+            " AND ".join(where),
+        )
+
+        cursor.execute(query, (app_name,))
+        columns = [desc[0] for desc in cursor.description]
```
jrgp: Could we simplify this to avoid the manual column name fetching + zip call by using a dict_cursor instead, like the other routes?

diegocepedaw: changed. dict_cursor wasn't playing nicely with mariadb which is why I had originally changed it

jrgp (approved):

PR #767, opened 2024-06-11
[DB SCHEMA CHANGE] Dynamic tracking notifications

diegocepedaw:
Allow for providing tracking notification destinations on incident creation time

Create a new dynamic_tracking_notification table to allow piggybacking off of the existing template/plan UI and still allow arbitrary tracking notifications to be specified on incident creation time. This will work similarly to how dynamic plans work today where the targets passed at incident creation time and are stored in a dynamic_plan_map table ato be later used to render the appropriate messages to the ad-hoc targets. 

To leverage this new feature a client would pass at incident creation, in addition to the context, a dynamic_tracking_notifications json field with the type of message and destinations for their dynamic tracking notifications. If a plan has dynamic_tracking enabled then the message processor will ignore the plan defined tracking notification and reuse the tracking template defined in the plan with the targets provided in dynamic_tracking_notifications. If the processor can’t find an appropriate template it will create an error message to expose the rendering failure to the end user like it currently exposes other messages that fail to send or render. In order to make this more explicit we will also add an extra `dynamic_tracking` tinyint(1) NOT NULL DEFAULT '0' column to the plan table and a checkbox in the plan editing UI so users can specify if they want to enable dynamic tracking messages for that plan

The contents of the dynamic_tracking_notifications table will be purged when the corresponding incidents are purged for retention.

### DIFF
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -7211,7 +7211,6 @@ class InternalIncidents():
         cursor = connection.cursor(db.dict_cursor)
         cursor.execute('''SELECT `id` FROM `incident` WHERE `active` = 1 AND `bucket_id` IN (SELECT `bucket_id` FROM `IMP_bucket_assignments` WHERE `node_id` = %s)''', node_id)
         result = cursor.fetchall()
-        print(node_id, result)
         cursor.close()
         connection.close()
         incident_ids = [row["id"] for row in result]
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -1187,6 +1187,7 @@ class Plan(object):
 
             cursor.execute(single_plan_query_tags, plan['id'])
             plan['tags'] = cursor.fetchall()
+            plan['dynamic_tracking'] = bool(plan.get('dynamic_tracking'))
 
             resp.body = ujson.dumps(plan)
             connection.close()
@@ -1444,6 +1445,7 @@ class Plans(object):
                     plan['tags'] = []
                 else:
                     plan['tags'] = ujson.loads(plan['tags'])
+            plan['dynamic_tracking'] = bool(plan.get('dynamic_tracking'))
             results.append(plan)
 
         if counts_only:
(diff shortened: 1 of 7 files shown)


src/iris/api.py
```diff
@@ -1999,13 +2016,54 @@ def on_post(self, req, resp):
...
+
+        dynamic_tracking_notifications = incident_params.get(
+            "dynamic_tracking_notifications", []
+        )
+        if not isinstance(dynamic_tracking_notifications, list):
+            raise falcon.HTTPBadRequest(
+                "Invalid Format", "dynamic_tracking_notifications should be a list."
+            )
+        for notification in dynamic_tracking_notifications:
+            if not isinstance(notification, dict):
```
jrgp, line 2046: At some point, using a library like marshmallow would help simplify this greatly by avoiding all of these manual checks

src/iris/api.py
```diff
@@ -2269,9 +2360,17 @@ def on_get(self, req, resp, incident_id):
 
             cursor.execute(single_incident_query_tags, incident['id'])
             incident['tags'] = cursor.fetchall()
-            connection.close()
 
             incident['context'] = ujson.loads(incident['context'])
+            # retrieve dynamic_tracking_notification for each incident
+            incident['dynamic_tracking'] = []
```
jrgp: Could this just be assigned to cursor.fetchall() like in the above examples, avoiding the for loop/etc?

src/iris/api.py
```diff
@@ -1179,6 +1187,7 @@ def on_get(self, req, resp, plan_id):
 
             cursor.execute(single_plan_query_tags, plan['id'])
             plan['tags'] = cursor.fetchall()
+            plan['dynamic_tracking'] = bool(plan.get('dynamic_tracking'))
```
jrgp: Nit: could we use [] instead of get() here, as the field is guaranteed to be in the response as it is in the select query?

src/iris/api.py

src/iris/api.py
diegocepedaw: yeah, we could definitely benefit massively from adding that style of validation to all endpoints

jrgp (approved):

PR #768, opened 2024-06-21
[DB SCHEMA CHANGE] restricted plans

diegocepedaw:
- add the ability to restrict plans to be modified only by Iris admins
- remove kazoo dependency from sender heartbeat

### DIFF
--- a/test/e2etest.py
+++ b/test/e2etest.py
@@ -1393,6 +1393,7 @@ def test_restricted_plan(sample_non_admin_user, sample_admin_user, sample_team, 
     # Test creating and deleting by ID
     re = requests.post(base_url + 'plans', json=data, headers=username_header(sample_non_admin_user))
     assert re.status_code == 401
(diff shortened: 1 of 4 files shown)


test/e2etest.py
```diff
@@ -1325,6 +1335,66 @@ def test_delete_plan(sample_user, sample_team, sample_template_name, sample_appl
...
+                    'template': sample_template_name
+                },
+            ],
+        ],
+        'isValid': True
+    }
+
+    # Test creating and deleting by ID
+    re = requests.post(base_url + 'plans', json=data, headers=username_header(sample_non_admin_user))
+    assert re.status_code == 401
```
jrgp, line 1395: Can we confirm the received error message is what we expect, to make sure we're getting a 401 for the intended reason?

diegocepedaw: added

jrgp (approved):

PR #769, opened 2024-08-02
add exists search operator for tag search

diegocepedaw:
add the ability to filter by resources that do or do not have a specific tag. E.g `/v0/incidents?tag_team__exists=true` would return only incidents that have the tag "team" regardless or the value. `/v0/incidents?tag_team__exists=false` would get incident that do not have a team tag.

### DIFF
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -62,6 +62,8 @@ operators = {
     'endswith': '%s LIKE CONCAT("%%%%", %s)',
 }
 
+exists_operator = "exists"
+
 
 def ts_to_sql_datetime(ts):
     return 'FROM_UNIXTIME(%s)' % ts
@@ -760,6 +762,8 @@ def gen_tag_where_subquery(connection, id_field, tag_table, resource_id, kwargs)
         # only process tag fields
         if not col.startswith(tag_prefix):
             continue
+        if op == exists_operator:
+            continue
         if op not in operators:
             raise HTTPBadRequest('invalid filter operator %s' % op)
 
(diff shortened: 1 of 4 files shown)


src/iris/api.py
```diff
@@ -3305,6 +3360,11 @@ def on_get(self, req, resp):
         if tag_subquery != "":
             where.append(tag_subquery)
 
+        tag_exists_subqueries = gen_tag_exists_where_subquery(connection, 'template_id', 'template_metadata_tag', '`template`.`id`', req.params)
+        if tag_exists_subqueries:
+            for tag_exists_subquery in tag_exists_subqueries:
+                where.append(tag_exists_subquery)
```
jrgp: I wonder if this could be simplified a bit if we created a new function that abstracts away both of the `gen_tag*` functions and returns a list of where clauses, so we can avoid maintaining both call sites 

Something like

```python
def gen_tag_wheres(...):
	wheres = []
	tag_subquery = gen_tag_where_subquery(...)
	if tag_subquery:
		wheres.append(tag_subquery)
	tag_exists_subqueries = gen_tag_exists_where_subquery(...)
	if tag_exists_subqueries:
		wheres.extend(tag_exists_subqueries)
    return wheres
```

And at call time in each of these uses, just:

```python
tag_wheres = gen_tag_wheres(connection, 'template_id', 'template_metadata_tag', '`template`.`id`', req.params)
if tag_wheres:
	where.extend(tag_wheres)
```

diegocepedaw: makes sense to me, added

src/iris/api.py
```diff
@@ -1418,9 +1475,10 @@ def on_get(self, req, resp):
         where += gen_where_filter_clause(
             connection, plan_filters, plan_filter_types, req.params)
 
-        tag_subquery = gen_tag_where_subquery(connection, 'plan_id', 'plan_metadata_tag', '`plan`.`id`', req.params)
-        if tag_subquery != "":
-            where.append(tag_subquery)
+        tag_subqueries = gen_tag_wheres(connection, 'plan_id', 'plan_metadata_tag', '`plan`.`id`', req.params)
+        if tag_subqueries:
+            for tag_subquery in tag_subqueries:
+                where.append(tag_subquery)
```
jrgp: Can avoid some boilerplate by replacing the for loop with where.extend()

jrgp (approved):

PR #770, opened 2024-09-11
allow sorting incident results by custom fields

diegocepedaw:
allows sorting incidents by specific fields. You can use the parameter `order_by` eg. `order_by=$FIELD` to sort the response by a specific field. Supported fields are:
        - `id`
        - `created`
        - `updated`
        - `owner`
        - `application`
        - `plan`
        - `context`
        - `active`
        
  Specify the order of the sort with the `order` parameter eg order=ASC. Supported values are:
        - ASC
        - DESC

### DIFF
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -1952,7 +1952,7 @@ class Incidents(object):
         req.params.pop('limit', None)
         target = req.get_param_as_list('target')
         req.params.pop('target', None)
-        order = req.get_param('order', default='DESC')
+        order = req.get_param('order', default=desc_order)
         req.params.pop('order', None)
         if order not in [asc_order, desc_order]:
             raise HTTPBadRequest('Invalid order parameter', 'Order parameter must be either "ASC" or "DESC"')
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -686,7 +686,7 @@ category_filter_types = {
 uuid4hex = re.compile(r'[0-9a-f]{32}\Z', re.I)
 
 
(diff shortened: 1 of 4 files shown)


src/iris/api.py
```diff
@@ -678,7 +696,8 @@ def stream_incidents_with_context(results, title=False):
                 row['title'] = row['context'].get(title_variable_name)
             else:
                 row['title'] = None
-        yield row
+        results_list.append(row)
+    return results_list
```
jrgp, line 700: With this change to this function, from being a generator yielding rows as it decodes them and instead accumulating them all at once in a list and return, it is no longer "streaming" the results and the name "stream_incidents_with_context" becomes misleading.

Can we either revert this change or rename the function?

src/iris/api.py
```diff
@@ -1988,8 +2032,13 @@ def on_get(self, req, resp):
         if where:
             query = query + ' WHERE ' + ' AND '.join(where)
 
+        if order_by is not None:
+            query += ' ORDER BY %s %s' % (incident_order_by_fields[order_by], order)
+        else:
+            query += ' ORDER BY `incident`.`created` DESC, `incident`.`id` DESC'
```
jrgp, line 2038: It looks like this always adds an ORDER BY clause even if one isn't requested. Depending on whether we have in index on the created field, this might be slower than expected. Would it make more sense to only order the result if it is explicitly requested?

src/iris/api.py
diegocepedaw: renamed

jrgp (approved):

PR #772, opened 2024-12-17
add configurable sso module flow

diegocepedaw:
Add configurable sso module flow that can assume user auth functions instead of the user/password auth. To use you must add your own sso module that takes in the request and returns a username if the auth headers exist and are valid and None if auth is unsuccessful. If sso auth fails user/password auth will still be available as backup

### DIFF
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -994,7 +994,11 @@ class AuthMiddleware(object):
             self.process_resource = self.debug_auth
 
     def debug_auth(self, req, resp, resource, params):
-        req.context['username'] = req.env.get('beaker.session', {}).get('user', None)
+        sso_username = self.sso_manager.authenticate(req)
+        if sso_username:
+            req.context['username'] = sso_username
+        else:
+            req.context['username'] = req.env.get('beaker.session', {}).get('user', None)
         method = req.method
 
         if resource.allow_read_no_auth and method == 'GET':
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -1952,7 +1952,7 @@ class Incidents(object):
         req.params.pop('limit', None)
         target = req.get_param_as_list('target')
         req.params.pop('target', None)
-        order = req.get_param('order', default='DESC')
+        order = req.get_param('order', default=desc_order)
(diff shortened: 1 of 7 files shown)


src/iris/sso.py
```diff
@@ -0,0 +1,34 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class SSOManager:
+
+    def __init__(self, config):
+        sso_cfg = config.get('sso', {})
+        if not sso_cfg.get('enabled', False):
+            logger.info('SSO passthrough enabled')
```
jrgp: Should this log say disabled instead of enabled?

src/iris/api.py
```diff
@@ -1038,7 +1042,13 @@ def debug_auth(self, req, resp, resource, params):
             return
 
     def process_resource(self, req, resp, resource, params):  # pragma: no cover
-        req.context['username'] = req.env.get('beaker.session', {}).get('user', None)
+        sso_username = self.sso_manager.authenticate(req)
```
jrgp, line 1045: Is this going to call the (expensive?) authenticate function on every page request? 

If so should this instead only call authenticate only once per session (upon login) and store the auth's username in the beaker session for fast access?

src/iris/sso.py
diegocepedaw: well the passthrough is enabled, but I think you're right that it would be less confusing to just say "SSO disabled"

src/iris/api.py
diegocepedaw: The check itself should be very fast. But more importantly I don't think we want to create a beaker session based off SSO login, I think we would want that login to work only when SSO auth is still fresh and since SSO does not have an active logout flow that pushes to our system we have no way of verifying that except going through the auth flow. Otherwise we can end up in an unexpected state of a user remaining logged in to Iris while logged out of SSO which I think we want to avoid since that bends the auth policy that SSO is trying to enforce.

jrgp (approved):

src/iris/api.py
jrgp: Makes sense, thanks for the clarification!

PR #775, opened 2025-03-19
Update CI to use Py3.10

hufman:
Updated Debian image and Python version to more recent supported versions. Also updated a few pip pins to support py3.10, and fixed up a few unit tests.


hufman:
Ah, I see this is not necessary on the active branch `experimental-sender`

PR #780, opened 2025-04-18
Refactor to support ONLY_FULL_GROUP_BY being enabled

diegocepedaw:
This PR introduces a set of refactors across the Iris API codebase to ensure compatibility with MySQL configurations where ONLY_FULL_GROUP_BY is enabled. This mode enforces stricter SQL standards, requiring all columns in the SELECT clause to be functionally dependent on the GROUP BY clause — which caused issues in several endpoints that relied on DISTINCT, aggregation, and partial groupings.

Enabling ONLY_FULL_GROUP_BY improves MySQL compliance and ensures queries are deterministic and optimizer-friendly. This change also sets the stage for moving to stricter SQL environments like managed cloud databases.

Testing:
- Verified behavior with ONLY_FULL_GROUP_BY enabled in dev DB
-  Confirmed correctness of tag filtering and metadata enrichment
- All e2e tests pass, including tag update flows, search queries, and grouped counts

### DIFF
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -1952,7 +1952,7 @@ class Incidents(object):
         req.params.pop('limit', None)
         target = req.get_param_as_list('target')
         req.params.pop('target', None)
-        order = req.get_param('order', default='DESC')
+        order = req.get_param('order', default=desc_order)
         req.params.pop('order', None)
         if order not in [asc_order, desc_order]:
             raise HTTPBadRequest('Invalid order parameter', 'Order parameter must be either "ASC" or "DESC"')
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -2226,10 +2226,12 @@ class Incidents(object):
 
             if num_dynamic > 0:
                 target_list = incident_params.get('dynamic_targets', [])
-                if num_dynamic != len(target_list):
+                if num_dynamic > len(target_list):
                     raise HTTPBadRequest('Invalid number of dynamic targets')
 
-                for dynamic_target in target_list:
+                for idx, dynamic_target in enumerate(target_list):
+                    if idx >= num_dynamic:
+                        break
                     target = session.execute('''SELECT `target_role`.`id` AS `role_id`, `target`.`id` AS `target_id`
                                                 FROM `target` JOIN `target_role`
                                                     ON `target_role`.`type_id` = `target`.`type_id`
--- a/src/iris/api.py
+++ b/src/iris/api.py
@@ -37,7 +37,7 @@ from iris.sender.quota import (get_application_quotas_query,
 from iris.utils import sanitize_unicode_dict
 from iris.vendors.iris_slack import iris_slack
 
-from . import app_stats, cache, client, db, ui, utils
+from . import app_stats, cache, client, db, ui, utils, sso
 from .config import load_config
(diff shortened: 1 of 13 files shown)


src/iris/api.py
```diff
@@ -766,6 +766,14 @@ def is_valid_tracking_settings(t, k, tpl):
     return True, None
 
 
+def get_order_expression(field: str) -> str:
```
aouyang1: type of response should be `Optional[str]`

src/iris/api.py
```diff
@@ -1371,6 +1367,83 @@ def on_delete(self, req, resp, plan_id):
         resp.body = '[]'
 
 
+# This function performs a grouped count across requested fields for templates, returning counts per value
+# for each requested field. It supports tag metadata and consolidates all SQL into a single query per field
+# to improve performance and eliminate ONLY_FULL_GROUP_BY errors. When only tag fields are requested, it performs
+# a separate total count since tags alone aren't reliable for row count estimation.
+def count_fields_for_plan_fields(cursor, where_clause, fields, allowed_columns, extra_joins=""):
```
aouyang1: can we add the appropriate type hints here?

src/iris/api.py
```diff
@@ -1465,9 +1538,14 @@ def on_get(self, req, resp):
         fields = [f for f in fields if f in plan_columns] if fields else None
         req.params.pop('fields', None)
         if not fields:
-            fields = plan_columns
+            fields = list(plan_columns.keys())
+        requested_fields = list(fields)
```
aouyang1, line 1466: we don't need to do that cast here since above it looks like fields will always be a list

src/iris/api.py
```diff
@@ -1825,6 +1910,87 @@ def on_post(self, req, resp):
...
+
+        final_query = " UNION ALL ".join(union_queries)
+        cursor.execute(final_query)
+        for row in cursor.fetchall():
+            field = row['field']
+            value = row['value']
+            count = row['group_count']
+            field_counts.setdefault(field, {})[value] = count
+        if field_counts:
+            total_count = sum(field_counts[next(iter(field_counts))].values())
```
aouyang1: we can wrap these into separate functions since these seem to be getting pretty large with consolidations for non_tag_fields, tags, and not non_tag_fields. For myself that's not familiar with this, I'm constantly checking to see if any of the current variables are populated from the previous clause which separate functions can disambiguate.

src/iris/api.py
```diff
@@ -1825,6 +1910,87 @@ def on_post(self, req, resp):
         resp.set_header('Location', '/plans/%s' % plan_id)
 
 
+def count_fields_for_incidents(cursor, where_clause, fields, allowed_columns, extra_joins=''):
```
aouyang1: we can also make these methods part of their respective classes e.g. Plans, Incidents, Templates unless we plan to use these elsewhere outside the context of the objects.

src/iris/api.py
diegocepedaw: It was done so that requested_fields is a shallow copy of the original field names, so later modifications to fields (like removing or reordering items) won’t affect requested_fields

src/iris/api.py
diegocepedaw: I refactored them into more modular functions for clarity

src/iris/api.py

src/iris/api.py

aouyang1 (approved):
LGTM!

